PR #2, opened 2020-06-23
Build the prototype of the user page that shows the need-help tasks

LeonardZ0731:
Create the user_profile.html file to build the skeleton of the user page. Create the user_profile_style.css file to modify the appearance of the file.
Create the user_profile_script.js file to pop up a notification box when the user clicks the delete button for a certain task. If the user click the OK button, then the task will be deleted.


src/main/webapp/user_profile_style.css
```diff
@@ -0,0 +1,158 @@
...
+#button-container-wrap {
+	margin-top: 20px;
+	margin-left: 2%;
+}
+
+#container {
+	display: inline;
+}
+
+#container .help-button {
```
swparr, line 57: Some of the styles here will be common across all pages, it would be worth thinking about creating a file for that and sharing

LeonardZ0731: After Mafe and Denzil commit their pages, we would set a meeting to discuss which style should be shared across all the pages. We would try to make the style for all pages nearly the same.

PR #4, opened 2020-06-24
Mafe home page prototype

poncem91:
Implement the UI of the app's home page using hard coded data for the prototype. Users can see listed tasks, they can filter tasks by categories, they can offer help with tasks, add new tasks, and go to their user page.


src/main/webapp/index.jsp
```diff
@@ -0,0 +1,234 @@
...
+          <nav>
+              <div id="userpage-icon">
+                  <a href="user_profile.html">
+                      <i class="fas fa-user-circle fa-3x" title="Go to User Page"></i>
+                  </a>
+              </div>
+          </nav>
+          <div id="title">
+              NeighborHelp
+          </div>
```
LeonardZ0731, line 23: For the title, I would recommend using an `<h1>` element for that.

src/main/webapp/index.jsp
```diff
@@ -0,0 +1,234 @@
...
+                  <div class="categories" id="category-pets" onclick="filterBy('pets')">PETS</div>
+                  <div class="categories" id="category-misc" onclick="filterBy('misc')">MISC</div>
+              </div>
+              <div id="add-task">
+                  <i class="fas fa-plus-circle" aria-hidden="true" id="addtaskbutton" title="Add Task" onclick="showModal()"></i>
+              </div>
+          </div>
+
+          <!--Hard Coded List of Tasks-->
+          <div id="tasks-list">
```
LeonardZ0731, line 44: For the user page, I used `<table>` to list all the tasks. We can discuss about whether we want to use `<table>` for tasks or use `<div>` instead.

src/main/webapp/index.jsp
poncem91: That's probably better, I'll change this!

src/main/webapp/index.jsp
poncem91: I was thinking about that too. I like using `<div>`s since it gives me more flexibility when aligning things, but we also might not need to display the tasks in the same way on both pages, since on the main page you only see the basic details, and on the users page it's more of an progress status display. I don't know haha, let's see what others say :)

PR #5, opened 2020-06-25
Implement the datastore feature for the user_profile page

LeonardZ0731:


src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        // Get the task detail from the form input
+        String taskDetail = "";
+        String input = request.getParameter("edit-content-input");
+        if (input != null) {
+            taskDetail = input;
+        }
+
+        // If the input is nonempty and valid, set the taskDetail value to the input value
```
poncem91, line 139: This is still pending, right? I wasn't sure if you meant to leave it unfinished for now or if you accidentally deleted some lines here. It might be good in the meantime to redirect users back to the userpage, right now it sends you to an empty blank page.

src/main/java/com/google/sps/task/Task.java
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.task;
+
+public final class Task {
```
poncem91, line 17: Looks good! I'll add the user id things in here!

src/main/webapp/user_profile.html
```diff
@@ -21,49 +21,94 @@ <h1 id="title">My Tasks</h1>
...
+                <td><button class="complete-task" onclick="completeTask(this)">
+                    <i class="fa fa-check"></i>
+                </button></td>
+                <td><button class="abandon-task" onclick="completeTask(this)">
+                    <i class="fa fa-times"></i>
                 </button></td>
             </tr>
         </table>
     </div>
+    <div id="createTaskModal">
```
poncem91, line 75: There might be a way of consolidating this into the same form, I'm not sure if it'll even be better or not... I'll look into it! It just seems like they're almost exactly the same minus the http method request, so it might be something to think about :)

src/main/webapp/user_profile_script.js
```diff
@@ -12,9 +12,145 @@
...
+    const request = new Request(queryURL, {method: "GET"});
+    const response = await fetch(request);
+    const info = await response.json();
+    console.log(info[0]);
+    return info[0];
+}
+
+async function deleteTask(keyString) {
+    const info = await getInfo(keyString);
+    if (info.status !== "OPEN") {
```
poncem91, line 26: It might be worth considering not including the delete or edit buttons at all for IN PROGRESS tasks, if they're going to be received with this message. But this works as well!

src/main/webapp/user_profile_script.js
```diff
@@ -12,9 +12,145 @@
...
+}
+
 function deleteRow(row) {
     if (confirm("Are you sure that you want to delete this already published task?")) {
         var rowIndex = row.parentNode.parentNode.rowIndex;
         document.getElementById("need-help").deleteRow(rowIndex);
     }
+}
+
+function deleteInProgress() {
```
poncem91, line 58: I don't think this function or the editInProgress function ever get used/called, since the same alert is used within the editTask and deleteTask functions

src/main/webapp/user_profile.html
```diff
@@ -21,49 +21,94 @@ <h1 id="title">My Tasks</h1>
...
                 <td>Leonard</td>
-                <td>OPEN</td>
-                <td><button class="edit-task"><i class="fa fa-edit"></i></button></td>
-				<td><button class="delete-task" onclick="deleteRow(this)">
-                    <i class="fa fa-trash-o"></i>
+                <td>100 pts</td>
+				<td><button class="complete-task" onclick="completeTask(this)">
+                    <i class="fa fa-check"></i>
+                </button></td>
+                <td><button class="abandon-task" onclick="completeTask(this)">
```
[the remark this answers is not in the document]
LeonardZ0731: I haven't implemented the abandon feature so I randomly chose a function as the onclick function for the prototype. I will implement the abandon feature in the next PR.

src/main/java/com/google/sps/servlets/MyNeedHelpTaskServlet.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates new task entity and fetch saved tasks. */
+@WebServlet("/mytasks/needhelp")
+public class MyNeedHelpTaskServlet extends HttpServlet {
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        // "Leonard" will be replaced with the user's accound id in the future
```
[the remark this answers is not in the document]
LeonardZ0731: Cool. 😄

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.sps.task.Task;
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates new task entity and fetch saved tasks. */
```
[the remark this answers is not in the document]
LeonardZ0731: Oh I just copy the all these imports with the header comment from TaskServlet.java file. I'm so sorry for that. I will change it in the next commit.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,78 @@
...
+@WebServlet("/tasks/info")
```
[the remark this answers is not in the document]
LeonardZ0731: One problem here. The request sent to `/tasks` url will be handled by `TaskServlet.java`, but for that servlet file, I hope that the`GET` request would be responded with a json string that includes all the stored tasks, instead of the information of a single task. So since I cannot assign two `doGet` method in one servlet file, I had to use a new url for getting the info.

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -0,0 +1,144 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates new task entity and fetch saved tasks. */
+@WebServlet("/tasks")
+public class TaskServlet extends HttpServlet {
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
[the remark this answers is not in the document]
LeonardZ0731: Ahh you can have a try. I think it may work with your web page implemented

src/main/webapp/user_profile_script.js
LeonardZ0731: So `editInProgress` and `deleteInProgress` are two functions that I prepare for the fake version of my user profile page. Now that I have implemented the real functionality, I will remove them in the next commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: For now I would rather keep those buttons, because for the OPEN tasks, there will be two columns, "EDIT" and "DELETE". It looks strange if the "IN PROGRESS" tasks miss the two columns. I will try to come up with a better way to warn people not pressing the two buttons for the "IN PROGRESS" task in the next commit.

poncem91: Sounds good!

src/main/webapp/user_profile_script.js
poncem91: ahhh that makes sense, sounds good!

src/main/webapp/user_profile.html
LeonardZ0731: Yep. This kind of code replication really annoys me but I haven't came up with a good idea. I may leave them for the prototype but think about how to fix this in the MVP.

src/main/java/com/google/sps/servlets/TaskServlet.java
LeonardZ0731: Yeah I am still working on the doPut method, but I will fill this out in the next commit.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,78 @@
...
+        long timestamp = (long) entity.getProperty("timestamp");
+        long reward = (long) entity.getProperty("reward");
+        String status = (String) entity.getProperty("status");
+        String owner = (String) entity.getProperty("Owner");
+        String helper = (String) entity.getProperty("Helper");
+        String address = (String) entity.getProperty("Address");
+
+        Task taskEntry = new Task(keyString, detail, timestamp, status, reward, owner, helper, address);
+        System.out.println(status.equals("OPEN"));
+        List<Task> result = new ArrayList<>();
```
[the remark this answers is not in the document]
LeonardZ0731: Sure. I found that my previous bug is not related to using a single task instead of an array. It's actually caused by missing an `await`. I improve this in the next commit.

src/main/webapp/user_profile.html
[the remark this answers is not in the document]
poncem91: Ah gotcha, sounds good :)!

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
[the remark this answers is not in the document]
poncem91: No worries! I just wanted to make sure I understood correctly :)

src/main/java/com/google/sps/servlets/EditTaskServlet.java
```diff
@@ -0,0 +1,105 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that handles the request for editing the details and rewards for a certain task. */
+@WebServlet("/tasks/edit")
```
poncem91, line 40: Having a separate servlet for this might be the better approach, but it might be worth considering also consolidating this one into the general tasksservlet and using the same form in the user_profile.html file to avoid repetition. It might not be a good idea though - I'm not sure if mixing edit with creating functionalities is ideal

poncem91: Otherwise this all looks good to me :)!

LeonardZ0731: I will try to combine the two servlet files after completing the prototype. Maybe I would add one more input parameter for the servlet file.

peap:
Can we rename the Java package in some PR? "sps" doesn't make sense anymore; it could be replaced with neighborgood, for example.

src/main/java/com/google/sps/task/Task.java
```diff
@@ -0,0 +1,45 @@
...
```
peap, line 17: It would be helpful to add a method in this class that understands how to make a Task entity from Datastore into an instance of this class. You can do this by adding another constructor that has a different signature, namely a Datastore entity instance. Then you don't have to update a bunch of call sites whenever you add a field to this model.

src/main/java/com/google/sps/task/Task.java
```diff
@@ -0,0 +1,45 @@
...
+  private final String detail;
+  private final String keyString;
+  private final long timestamp;
```
peap, line 20: What time does this timestamp refer to? Can we give it a name to clarifies this?

src/main/java/com/google/sps/servlets/MyTaskServlet.java
```diff
@@ -0,0 +1,71 @@
...
+      String detail = (String) entity.getProperty("detail");
+      long timestamp = (long) entity.getProperty("timestamp");
+      long reward = (long) entity.getProperty("reward");
+      String status = (String) entity.getProperty("status");
+      String owner = (String) entity.getProperty("Owner");
+      String helper = (String) entity.getProperty("Helper");
+      String address = (String) entity.getProperty("Address");
+
+      Task taskEntry =
+          new Task(keyString, detail, timestamp, status, reward, owner, helper, address);
```
peap, line 62: See my comment about making this easier by moving this logic in `Task.java`.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,97 @@
...
+
+    String detail = (String) entity.getProperty("detail");
+    long timestamp = (long) entity.getProperty("timestamp");
+    long reward = (long) entity.getProperty("reward");
+    String status = (String) entity.getProperty("status");
+    String owner = (String) entity.getProperty("Owner");
+    String helper = (String) entity.getProperty("Helper");
+    String address = (String) entity.getProperty("Address");
+
+    Task taskEntry = new Task(keyString, detail, timestamp, status, reward, owner, helper, address);
```
peap, line 62: See my comment about making this easier by moving this logic in `Task.java`.

src/main/java/com/google/sps/servlets/MyTaskServlet.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that fetch all saved tasks whose owner is the current user. */
+@WebServlet("/mytasks")
+public class MyTaskServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // "Leonard" will be replaced with the user's accound id in the future
+    String keyword = request.getParameter("keyword");
+    Filter ownerFilter = new FilterPredicate(keyword, FilterOperator.EQUAL, "Leonard");
```
peap, line 43: We'll also be filtering on the "owner" field, right? Should we be doing that already?

src/main/java/com/google/sps/servlets/TaskServlet.java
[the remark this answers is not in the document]
peap: I think we will want this simplification at some point, yes. It looks a bit redundant to see TaskServlet, TaskInfoServlet, EditTaskServlet, and MyTaskServlet all in the same directory.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,97 @@
...
+    Key taskKey = KeyFactory.stringToKey(keyString);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity entity;
+    try {
+      entity = datastore.get(taskKey);
+    } catch (EntityNotFoundException e) {
+      System.err.println("Unable to find the entity based on the input key");
+      response.sendError(
+          HttpServletResponse.SC_NOT_FOUND, "The entity is not found in the database");
```
peap, line 50: Same comment here about using language that makes sense to the end user.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,97 @@
...
+    String detail = (String) entity.getProperty("detail");
+    long timestamp = (long) entity.getProperty("timestamp");
+    long reward = (long) entity.getProperty("reward");
+    String status = (String) entity.getProperty("status");
+    String owner = (String) entity.getProperty("Owner");
+    String helper = (String) entity.getProperty("Helper");
+    String address = (String) entity.getProperty("Address");
+
+    Task taskEntry = new Task(keyString, detail, timestamp, status, reward, owner, helper, address);
+    System.out.println(status.equals("OPEN"));
```
peap, line 63: In the future, let's remove ad hoc debugging print-statements like this before committing. There can be huge value in adding targeted logging statements that show interesting information and context, but a statement like this would end up filling the logs of your deployed application with a bunch of "true" and "false" lines, which will just make it harder to find the interesting logs.

This is a perfectly valid thing to do for local development and debugging, but we should hold the code we check in to a higher standard.

LeonardZ0731:
I will rename the package to neighborgood in a separate PR. (I actually don't know what does sps mean, but the walkthrough examples all name the java package as sps, so I just copy that for the java package)

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
```diff
@@ -0,0 +1,97 @@
...
+
+    Gson gson = new Gson();
+    String json = gson.toJson(taskEntry);
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
peap, line 72: This post method is quite similar to the post method in EditTaskServlet. Can we combine the two?

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -0,0 +1,122 @@
...
+        // Serve the GET request sent from home page to fetch all the tasks
+        return;
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        // Get the rewarding points from the form
+        int rewardPts = getRewardingPoints(request, "reward-input");
```
peap, line 51: Why not make the fields use the same name in both HTML forms?

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -0,0 +1,122 @@
...
+
+        // Get the task detail from the form input
+        String taskDetail = "";
+        String input = request.getParameter("task-content-input");
+        if (input != null) {
+            taskDetail = input;
+        }
+
+        // If the input is nonempty and valid, set the taskDetail value to the input value
+        if (!taskDetail.equals("")) {
```
peap, line 66: What happens if there are no details provided?

src/main/webapp/user_profile.html
```diff
@@ -115,5 +49,25 @@ <h1>CREATE A NEW TASK: </h1>
             </form>
         </div>
     </div>
+    <div id="editTaskModal">
+        <div id="editModal">
```
peap, line 53: editModal smells more like a class than an ID; will there be other entities that users can edit within a modal?

src/main/webapp/user_profile.html
```diff
@@ -115,5 +49,25 @@ <h1>CREATE A NEW TASK: </h1>
...
+            <span id="edit-close-button" onclick="closeEditModal()">&times;</span>
+            <form id="edit-task-form" action="/tasks/edit" method="POST">
+                <h1>EDIT THE CURRENT TASK: </h1>
+                <div>
+                    <label for="edit-content-input">Task Detail:</label>
+                    <br/>
+                </div>
+                <textarea name="edit-content-input" id="edit-content-input"></textarea>
```
peap, line 61: "edit-detail-input"?

src/main/webapp/user_profile.html
```diff
@@ -115,5 +49,25 @@ <h1>CREATE A NEW TASK: </h1>
...
+                <br/>
+                <label for="edit-point-input">Rewarding Points: </label>
+                <input type="number" id="edit-point-input" name="edit-input" min="0" max="200">
```
peap, line 64: "edit-input" seems like a generic name for this part of the form; how about "edit-point-input", to match the ID?

src/main/webapp/user_profile.html
```diff
@@ -115,5 +49,25 @@ <h1>CREATE A NEW TASK: </h1>
...
+                <textarea name="task-id" id="task-id-input"></textarea>
```
peap, line 65: This would be better as `<input type="hidden">`, which is meant for passing along invisible data in a form.

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-function deleteRow(row) {
-    if (confirm("Are you sure that you want to delete this already published task?")) {
-        var rowIndex = row.parentNode.parentNode.rowIndex;
-        document.getElementById("need-help").deleteRow(rowIndex);
+async function getInfo(keyString) {
```
peap, line 15: "getTaskInfo" would better describe what this function does.

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
...
+
+async function deleteTask(keyString) {
+    const info = await getInfo(keyString);
+    if (info.status !== "OPEN") {
+        window.alert("You can only delete an 'OPEN' task.")
+    } else {
+        if (confirm("Are you sure that you want to delete the task?")) {
+            const queryURL = "/tasks?key=" + keyString;
+            const request = new Request(queryURL, {method: "DELETE"});
+            const response = await fetch(request);
```
peap, line 32: Should we hide the row after the delete succeeds? Right now the task stays on the page until I refresh.

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
...
+    console.log(info);
+    if (info.status !== "OPEN") {
+        window.alert("You can only edit an 'OPEN' task.")
+    } else {
+        document.getElementById("edit-content-input").value = info.detail;
+        document.getElementById("edit-point-input").value = info.reward.toString();
+        const id_input = document.getElementById("task-id-input");
+        id_input.value = info.keyString;
+        id_input.readOnly = true;
+        id_input.style.display = "none"
```
peap, line 48: These two properties don't need to be explicitly set if you use `<input type="hidden">`.

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
...
-        document.getElementById("offer-help").deleteRow(rowIndex);
+async function completeTask(keyString) {
+    const info = await getInfo(keyString);
+    if (info.status !== "IN PROGRESS") {
+        window.alert("You have already marked the task as complete.");
+    } else {
+        if (confirm("Are you sure that you have already completed the task?")) {
+            const queryURL = "/tasks/info?key=" + keyString + "&status=" + "COMPLETE: AWAIT VERIFICATION";
+            const request = new Request(queryURL, {method: "POST"});
+            const response = await fetch(request);
```
peap, line 61: Same here: should we update something visually once this is marked as done?

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
...
+        id_input.style.display = "none"
+        document.getElementById("editTaskModal").style.display = "block";
     }
 }
 
-function completeTask(row) {
-    if (confirm("Are you sure that you have already completed the task?")) {
-        var rowIndex = row.parentNode.parentNode.rowIndex;
-        document.getElementById("offer-help").deleteRow(rowIndex);
+async function completeTask(keyString) {
```
peap, line 53: I'm not sure how to create a task locally as one user and look at it as another user. As a result, I don't know how to get my local instance into a state where I can try assigning a task to myself and marking it complete. Do you have any advice for how to test this?

src/main/webapp/user_profile_script.js
```diff
@@ -12,17 +12,67 @@
...
+
+async function abandonTask(keyString) {
+    const info = await getInfo(keyString);
+    if (info.status !== "IN PROGRESS") {
+        window.alert("You have already marked the task as complete.");
+    } else {
+        if (confirm("Are you sure that you want to abandon the task?")) {
+            const queryURL = "/tasks/info?key=" + keyString + "&status=" + "OPEN";
+            const request = new Request(queryURL, {method: "POST"});
+            const response = await fetch(request);
```
peap, line 74: And here: maybe a refresh-row function would be useful.

src/main/webapp/user_profile_script.js
```diff
@@ -52,10 +104,105 @@ function closeModal() {
...
         modal.style.display = "none";
     }
+    var editModal = document.getElementById("editTaskModal");
+    if (event.target == editModal) {
+        editModal.style.display = "none";
+    }
+}
+
+async function displayNeedHelpTasks() {
+    const queryURL = "mytasks?keyword=Owner";
```
peap, line 125: I'd recommend using an absolute URL, almost always: `/mytasks?...`

src/main/webapp/user_profile_script.js
```diff
@@ -52,10 +104,105 @@ function closeModal() {
...
+    const taskResponse = await response.json();
+    const needHelpTable = document.getElementById("need-help");
+    needHelpTable.innerHTML = "";
+    var headerRow = document.createElement("tr");
+    var headers = ["Task Overview", "Helper", "Status", "Edit", "Delete"];
+    for (var index = 0; index < headers.length; index++) {
+        var th = document.createElement("th");
+        th.appendChild(document.createTextNode(headers[index]));
+        headerRow.appendChild(th);
+    }
```
peap, line 137: I'd prefer this static HTML to live in the HTML file; you can use JS to hide/show it as needed.

src/main/webapp/user_profile_script.js
```diff
@@ -52,10 +104,105 @@ function closeModal() {
...
+        editTd.appendChild(editBtn);
+        var deleteTd = document.createElement("td");
+        var deleteBtn = document.createElement("button");
+        deleteBtn.className = "delete-task";
+        deleteBtn.addEventListener("click", function () { deleteTask(keyStringCopy) });
+        deleteBtn.innerHTML = (task.status === "OPEN") ? '<i class="fa fa-trash-o"></i>':'<i class="fa fa-ban"></i>';
+        deleteTd.appendChild(deleteBtn);
+        tr.appendChild(editTd);
+        tr.appendChild(deleteTd);
+        needHelpTable.appendChild(tr);
```
peap, line 163: Something to consider: you can make a template row in your HTML file that's hidden, then this function can clone it, replace all the relevant fields with Task-specific information, then append it to the table. That should simplify and shorten this function a bit.

src/main/webapp/user_profile_script.js
```diff
@@ -52,10 +104,105 @@ function closeModal() {
...
+    for (var index = 0; index < taskResponse.length; index++) {
+        var tr = document.createElement("tr");
+        var task = taskResponse[index];
+        var data = [task.detail, task.helper, task.status];
+        for (var i = 0; i < data.length; i++) {
+            var td = document.createElement("td");
+            td.appendChild(document.createTextNode(data[i]));
+            tr.appendChild(td);
+        }
+        const keyStringCopy = task.keyString.slice();
```
peap, line 148: What's the purpose of this copy?

src/main/webapp/user_profile_script.js
```diff
@@ -52,10 +104,105 @@ function closeModal() {
...
+        deleteBtn.innerHTML = (task.status === "OPEN") ? '<i class="fa fa-trash-o"></i>':'<i class="fa fa-ban"></i>';
+        deleteTd.appendChild(deleteBtn);
+        tr.appendChild(editTd);
+        tr.appendChild(deleteTd);
+        needHelpTable.appendChild(tr);
+    }
+}
+
+async function displayOfferHelpTasks() {
+    const queryURL = "mytasks?keyword=Helper";
```
peap, line 168: Prefer an absolute URL: `/mytasks?...`

src/main/webapp/user_profile_style.css
```diff
@@ -149,15 +149,15 @@ th, td {
...
 	background-color: red;
 }
 
-.delete-task:hover i {
+.delete-task:hover i, .abandon-task:hover i {
 	color: white;
 }
 
-#createTaskModal {
+#createTaskModal, #editTaskModal {
```
peap, line 160: Instead of slowly building up a list of IDs, let's make this a class and apply it to the elements. Same for the next several rules that change in this PR.

src/main/webapp/user_profile_script.js
LeonardZ0731: At first I didn't make a copy for the keyString. But then I found that for all the onclick functions I set for the edit buttons, when the button is clicked, the function editTask(keyString) should be called and the keyString should be different for each onclick function. However, if I don't make a copy of the keyString, all the editTask function will take in the same keyString input. Thus, it seems that without making a copy, all the input keyString would just be aliases referring to the same keyString. As a result, I make a copy for each separate keyString.

src/main/java/com/google/sps/servlets/EditTaskServlet.java
```diff
@@ -0,0 +1,105 @@
...
+        // If the input is nonempty and valid, set the taskDetail value to the input value
+        if (!taskDetail.equals("")) {
+            Key taskKey = KeyFactory.stringToKey(keyString);
+            DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+            Entity task;
+            try {
+                task = datastore.get(taskKey);
+            } catch (EntityNotFoundException e) {
+                System.err.println("Unable to find the entity based on the input key");
+                response.sendError(HttpServletResponse.SC_NOT_FOUND, "The entity is not found in the database");
```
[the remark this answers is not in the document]
LeonardZ0731: Sure. I would modify the error message sent by `EditTaskServlet.java` and `TaskInfoServlet.java` in the new commit.

src/main/java/com/google/sps/servlets/EditTaskServlet.java
```diff
@@ -0,0 +1,105 @@
...
+
+        // Get the task detail from the form input
+        String taskDetail = "";
+        String input = request.getParameter("edit-content-input");
+        if (input != null) {
+            taskDetail = input;
+        }
+
+        // If the input is nonempty and valid, set the taskDetail value to the input value
+        if (!taskDetail.equals("")) {
```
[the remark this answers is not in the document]
LeonardZ0731: When the edit task form is created, the task detail and original rewarding points are automatically filled in the form. Then the user can modify the task detail and reward point if they want. So if the user only update the rewarding points without updating the task details, when he submit the form, the servlet file can still get the original task detail. So in this way the user could update reward points without updating the task details. But if the user remove the task detail, I would reject the edit request and redirect him back to the user profile page.

src/main/java/com/google/sps/task/Task.java
LeonardZ0731: Ahh that really helps for entities with so many fields. I would do this in the new commit by adding a new constructor that takes in an Entity and outputs a task instance.

src/main/java/com/google/sps/servlets/MyTaskServlet.java
LeonardZ0731: I replaced this block of code with a single line of code `myTasks.add(new Task(entity));` in the newest commit.

src/main/java/com/google/sps/task/Task.java
LeonardZ0731: Oh this timestamp should refer to the time that the task was created. I would give it a name `creationTime` in the next commit.

LeonardZ0731: For this PR, I didn't modify the property name "timestamp" set for the task entity in datastore because I have already stored a lot of fake task entities with property name "timestamp". But in the next PR, I will modify the property name to "creationTime".

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
LeonardZ0731: In the new commit I replaced this block of code with a single line `Task taskEntry = new Task(entity)`.

src/main/java/com/google/sps/servlets/MyTaskServlet.java
LeonardZ0731: Ahh actually it's my naming problem. The filter is actually different depending on the input keyword. The doGet function mainly serves the GET request sent from the Need Help page and the Offer Help page. The Need Help play displays all the tasks whose owner is current user so the GET request sent by that page will have a keyword of "Owner". Then the filter will filter on the "owner" field. For the Offer help page, it will send a GET request with keyword "Helper". Thus the filter will filter on the "helper" field. So I would rename the `Filter` element in the new commit.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
LeonardZ0731: Sure. I always forgot to remove these debugging print statements after finding the bugs. I would remember to do that in the future. I will remove all the debugging print statements in the next commit.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
LeonardZ0731: Sure. I changed the response to "The requested task could not be found" in the new commit.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
LeonardZ0731: Emm I don't think we can combine the two functions because what the two post method modify are really different. For the doPost method in `TaskInfoServlet.java`, it modifies the status of the task. It will change the task status to "OPEN" or "AWAIT VERIFICATION" and change the task Helper field accordingly. However, the doPost method in `EditTaskServlet.java` modifies the rewarding points and task details. So the field that the two methods modify are distinct and maybe combing them has little benefit. So I would prefer not combining them for now. 

I may consider adding a post method that modifies all the fields of a task entity in a new PR. After implementing that post method, I will combine the two post methods.

src/main/java/com/google/sps/servlets/TaskServlet.java
LeonardZ0731: I actually wanted to reject the request if the task detail is not provided. So at that time, I just didn't handle the request and redirected the user back to the user profile page. Now I think maybe respond with a 403 Forbidden error can be a better idea. I will make the change in the new commit.

src/main/webapp/user_profile.html
LeonardZ0731: Ahh I actually called "Task Detail" as "Task Content" when I first wrote the code and after changing the name to "Task Detail" I forgot to modify the name for the input textareas. I would fix it in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Actually I didn't consider this kind of situation when I wrote the code 😭. So I don't have any good advice on how to test this. But maybe we could prevent this kind of situation by not allowing the user to assign the task to a certain people. Instead, users have to undertake tasks on the main page, and when displaying tasks on the main page, we will filter out all the tasks that are published by the current user. In this way, a user can never complete a task that is published by himself. 

Apart from this, I modify the `completeTask` function in the new commit such that before changing the status of the task, it would first check the owner and the helper of the task. If they are the same user, it would send an alert and then delete the task instead of changing the status to "COMPLETE: AWAIT VERIFICATION". I think this could be a good solution to this edge case.

src/main/java/com/google/sps/servlets/TaskServlet.java
LeonardZ0731: Oh I don't know we can have the same name for a input text box across different forms. I will use the same name in the new commit.

src/main/webapp/user_profile.html
LeonardZ0731: I change the name to "reward-input" to match the name of the create task form in the new commit.

src/main/webapp/user_profile.html
LeonardZ0731: Okay. I would replace the task id input `textarea` with a hidden input in the next commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure. I will change the function name in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Okay. After changing the task ID input from `textarea` to hidden input, I removed the two lines in the new commit.

src/main/webapp/user_profile.html
LeonardZ0731: Emm I don't think the users edit other entities within the modal. The modal should only serve to edit tasks. Thus, I rename the ids of the two `<div>` element in the new commit. The outer `<div>` is now named as `editTaskModalWrapper` and the inner `<div>` element is now named as `editTaskModal`.

src/main/webapp/user_profile_script.js
LeonardZ0731: In the new commit, I force the page to load the need help table again after the deletion occurs. Now the task should be removed once the user deleted the task.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure. I do the same here in the new commit and the offer help table is now forced to load again after the user mark a task as complete or abandon the task.

src/main/webapp/user_profile_script.js
LeonardZ0731: I force the offer help table to load again here in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: The reason why I add a header with js code every time is that I haven't found a proper way to remove the other rows without removing the first row. The only way I knew that can remove multiple rows in one step is to set the innerHtml to empty string. But in this way, the static header will also be removed. So I add the header rows dynamically here. I will try to find a way to use static HTML in a new PR.

src/main/webapp/user_profile_script.js
LeonardZ0731: Oh the suggestion is really helpful. It can make the two functions `displayNeedHelpTasks()` and `displayOfferHelpTasks()` much more concise. I will try this way in a new PR.

src/main/webapp/user_profile_script.js
LeonardZ0731: I forgot to add the slash when I change the url that handles the POST request. I will add it in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure.  I fix this in the new commit.

src/main/webapp/user_profile_style.css
LeonardZ0731: Sure. I let the two modal wrappers belong to the same class `modalWrapper` in the new commit. Similarly, I let the two modals, close buttons of the two modals, and two task tables have their own class respectively. Then I modify the css file to accommodate the change in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: I just found a way to replace the innerhtml of a table to the inner html of the first row of that table. Now I am able to use a static row header HTML instead of adding the heading row each time I display the table. I will make the change in the new commit.

peap:
> I will rename the package to neighborgood in a separate PR. (I actually don't know what does sps mean, but the walkthrough examples all name the java package as sps, so I just copy that for the java package)

Sounds good, thanks! "sps" refers to the project that this summer's STEP project is based on, Software Product Sprint (https://buildyourfuture.withgoogle.com/programs/softwareproductsprint/).

src/main/java/com/google/sps/task/Task.java
peap: Sounds good, thanks.

src/main/java/com/google/sps/servlets/TaskInfoServlet.java
peap: OK, sounds good.

I was thinking there could be a general-purpose method that updates only the fields of a Task that are present in the HTTP request, rather than requiring all of them to be sent in each request.

peap: Doing any combination in a separate PR sounds ok though.

src/main/java/com/google/sps/servlets/TaskServlet.java
peap: Yep, the same `name` can appear in many places on an HTML page, though it gets complicated if multiple belong to the same `form`. It's just the `id` that needs to be unique on the page.

src/main/java/com/google/sps/servlets/TaskServlet.java
peap: [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) is probably more appropriate; 403 Forbidden usually indicates a permissions error.

src/main/webapp/user_profile.html
peap: I don't mean that users will edit another kind of entity in *this* modal, just that you might have other entities in the future that you want to let users edit in a modal. At that time, it would be handy if this were a class.

I see that you've made classes now, so thanks! (`modalWrapper` and `modal`)

src/main/webapp/user_profile_script.js
peap: Sure, sounds good.

src/main/java/com/google/sps/servlets/TaskServlet.java
LeonardZ0731: Thanks! I will change it in a new commit.

src/main/webapp/user_profile_script.js
peap: Sorry, I just meant that I can't see how to log in as a different user to try and assign some tasks to that user. But with the future PRs in this repo that I've now seen, hopefully that will be possible soon.

How have you been testing the "Offer help" side of this page, including completing tasks and such?

src/main/webapp/user_profile_script.js
peap: Ah, yes it probably has to do with the anonymous function on line 155 that was referring to the loop variable `index`. Keeping this copy SGTM.

src/main/java/com/google/sps/servlets/EditTaskServlet.java
```diff
@@ -39,67 +29,74 @@
...
-            Entity task;
-            try {
-                task = datastore.get(taskKey);
-            } catch (EntityNotFoundException e) {
-                System.err.println("Unable to find the entity based on the input key");
-                response.sendError(HttpServletResponse.SC_NOT_FOUND, "The entity is not found in the database");
-                return;
-            }
+    // If the input is nonempty and valid, set the taskDetail value to the input value
+    if (!taskDetail.equals("")) {
```
peap, line 50: I think it's simpler and easier to follow this method if we return early (with an error?) when taskDetail is empty, like we do above if rewardPts is -1. Then you can dedent the rest of the method body to indicate that it's the normal flow of the method. Pseudocode:

```
...
if rewardPts == -1 {
  return error
}
...
if taskDetail.equals("") {
  return error
}
... do all the rest of the things ...

File: src/main/java/com/google/sps/servlets/TaskServlet.java [hunk 42]
```diff
@@ -39,84 +28,90 @@
...
\ No newline at end of file
+  @Override
+  public void doDelete(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    String keyString = request.getParameter("key");
+
+    Key taskKey = KeyFactory.stringToKey(keyString);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.delete(taskKey);
```
peap, line 112: Will you leave a TODO comment here for error handling? We should handle the case where the user attempts to delete a non-existent task.

src/main/webapp/user_profile_script.js
```diff
@@ -95,47 +106,39 @@ function showOfferHelp() {
...
-    needHelpTable.innerHTML = "";
-    var headerRow = document.createElement("tr");
-    var headers = ["Task Overview", "Helper", "Status", "Edit", "Delete"];
-    for (var index = 0; index < headers.length; index++) {
-        var th = document.createElement("th");
-        th.appendChild(document.createTextNode(headers[index]));
-        headerRow.appendChild(th);
-    }
-    needHelpTable.appendChild(headerRow);
+    needHelpTable.innerHTML = needHelpTable.rows[0].innerHTML;
```
peap, line 141: I don't think this is doing quite what we want, semantically. Here's the HTML I see on my local instance:

```
<table class="task-table" id="need-help" style="display: table;">
    <tbody>
        <tr>
            <th>Task Overview</th>
            <th>Helper</th>
            <th>Status</th>
            <th>Edit</th>
            <th>Delete</th>
        </tr>
    </tbody>
    <tr>
        <td>this is my task</td>
        <td>N/A</td>
        <td>OPEN</td>
        <td><button class="edit-task"><i class="fa fa-edit"></i></button></td><td><button class="delete-task"><i class="fa fa-trash-o"></i></button></td>
    </tr>
</table>
```

See how the table has a `tbody` child with the header rows, then a `tr` element afterwards? Better would be to wrap the header row in a `thead` element that is a child of the `table` element, then have this JS stick to modifying just the `tbody`, which should contain all the Task data rows.

src/main/webapp/user_profile.html
```diff
@@ -28,19 +28,36 @@ <h1 id="title">My Tasks</h1>
     </div>
     <br/>
     <div id="task-list">
-        <table id="need-help"></table>
-        <table id="offer-help"></table>
+        <table class="task-table" id="need-help">
+            <tr>
```
peap, line 32: Header rows should belong to a `thead` element. You can also pre-populate this table with an empty `tbody` element for the data rows to be added.

src/main/webapp/user_profile_script.js
```diff
@@ -12,66 +12,77 @@
...
     if (info.status !== "IN PROGRESS") {
         window.alert("You have already marked the task as complete.");
+    } else if (info.owner === info.helper) {
+        window.alert("You cannot complete a task published by yourself! The task will now be removed from the system!");
+        var queryURL = "/tasks/info?key=" + keyString + "&status=" + "OPEN";
+        var request = new Request(queryURL, {method: "POST"});
+        var response = await fetch(request);
+        queryURL = "/tasks?key=" + keyString;
+        request = new Request(queryURL, {method: "DELETE"});
+        response = await fetch(request);
```
peap, line 64: It seems rather harsh to delete the task if the user manages to try marking it complete. Shouldn't we just reject the attempt, display the first half of the alert, and call it good?

src/main/java/com/google/sps/servlets/EditTaskServlet.java
LeonardZ0731: Sure. I will change it in a new commit for both the `EditTaskServlet.java` and `TaskServlet.java` files.

src/main/java/com/google/sps/servlets/TaskServlet.java
LeonardZ0731: Oh sure, I will add a TODO comment here and handle the exceptional case in a new PR.

src/main/webapp/user_profile_script.js
LeonardZ0731: Ah using `thead` and `tbody` make things much easier. I put the header row in a `thead` element in the new commit and empty the `tbody` each time I load the table.

src/main/webapp/user_profile.html
LeonardZ0731: Sure. In the new commit the header row is put in a `thead` element and each time I load the table, I first empty the `tbody` element. Then appending `tr` element to `tbody` one by one.

src/main/webapp/user_profile_script.js
LeonardZ0731: Ah yeah. I could do that, making the user have no choice but to abandon the task at the end. I will fix this in the new commit.

PR #7, opened 2020-06-25
fix bug in homepage prototype

poncem91:
Fix bug that allowed guest users to see the go-to-userpage button, as well as a bug that threw an error if a user tried logging out, back in, and logging out again (due to a typo in the function name).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.sps</groupId>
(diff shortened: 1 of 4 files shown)


poncem91:
> It's good practice to have each of these fixes in a separate PR. It seems like additional work, but it prevents any feedback on one change from holding back the push of the other. No need to change for this PR but having multiple, small PR's is preferred.

My bad! I'll do this moving forward, thanks!

PR #11, opened 2020-06-26
Denzil Admin Prototype

denzilbilson-google:


src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+					<li><a>My Account</a></li>
+					<li><a>Statistics</a></li>
+					<li><a>Settings</a></li>
+					<li><a>Logout</a></li>
+				</ul>
+			</nav>
+
+			<main>
+				<h1>Admin Prototype</h1>
+				<div class="notification-grid">
```
poncem91, line 25: We might want to think whether we should remove this for the prototype and re-add it if we end up having a notification system. My understanding is that the prototype includes what we are going to implement, but I could be wrong!

src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+						This thing just happened.
+					</div>
+					<div>
+						<h3>Notification 3</h3>
+						This thing just happened.
+					</div>
+				</div>
+
+				<div class="flex-grid">
+                    <div id="card-container">
```
poncem91, line 41: This looks awesome! 👍

src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+		<script src='https://kit.fontawesome.com/a076d05399.js'></script>
+        <script src="https://www.gstatic.com/charts/loader.js"></script>
+        <script src="admin_dashboard_script.js"></script>
+	</head>    
+	<body onresize="drawChart()">
+		<div class="wrapper">
+			<nav>
+				<header>NeighborGood</header>
+				<ul>
+					<li><a class="active">Dashboard</a></li>
```
poncem91, line 15: This looks good! You might want to think though what you want these to do when clicked and add some 'fake functionality' for the prototype :)

target/NeighborHelp-1/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
+<!DOCTYPE html>
```
poncem91, line 1: This is probably alright for now, but you might want to add a .gitignore file to ignore the /target folder when doing commits.  Checkout the one that's currently on the github master branch, if you want you can just copy that into your /Neighborgood folder.

denzilbilson-google: Sounds like a swell idea!

src/main/webapp/admin_dashboard_script.js
```diff
@@ -0,0 +1,37 @@
+google.charts.load('current', {'packages':['line']});
+google.charts.setOnLoadCallback(drawChart);
+
+document.getElementsByTagName("body").addEventListener("onresize", drawChart);
```
peap, line 4: I get this error locally:

```
Uncaught TypeError: document.getElementsByTagName(...).addEventListener is not a function
    at admin_dashboard_script.js:4
```

Note that `getElementsByTagName` returns an `HTMLCollection`, not an `HTMLElement`, so `addEventListener` isn't valid. Since there's only one `body`, though, you could use `document.querySelector` to get the single element back, then add the event listener.

src/main/webapp/admin_dashboard_script.js
```diff
@@ -0,0 +1,37 @@
...
+function _init_(){
+}
```
peap, line 6: Delete this until you need it.

src/main/webapp/admin_dashboard_style.css
```diff
@@ -0,0 +1,192 @@
...
+	list-style: none;
+}
+.wrapper {
+	display: flex;
+	min-height: 100%;
+}
+nav {
+	width: 200px;
+	background: #eee;
+    border-right: 1px solid #222;
```
peap, line 28: Could you run this file through a CSS formatter? It's nice to see consistent indentation, spaces, etc. There are online tools you can use, Maven plugins, etc.; maybe cloud shell editor can do it?

I think the main issue is that there's a mixture of tabs and spaces; if your tab width is set to 4, you might not notice but it's quite apparent in the GitHub UI. Maybe a search-and-replace of tabs for 4 spaces would fix matters.

src/main/webapp/admin_dashboard.html
peap: Agreed! Looks great, and it would be nice to see what you intend these sections to do. (That should probably be a follow-up PR, though.)

src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+						This thing just happened.
+					</div>
+					<div>
+						<h3>Notification 3</h3>
+						This thing just happened.
+					</div>
+				</div>
+
+				<div class="flex-grid">
+                    <div id="card-container">
```
peap: Same comment as in the CSS file; could you run this through an HTML formatter? The mixture of tabs and spaces makes the file hard to read if you don't have tabs set to display as four spaces.

src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+                        <h2>Tasks</h2>
+                        <div id="card-div">
+                            <ul class="tasks">
```
peap: I understand that this is fake prototype data, but just to be sure: will this section show user tasks from neighbors in the app, or meta-tasks about the website itself?

src/main/webapp/admin_dashboard.html
```diff
@@ -0,0 +1,111 @@
...
+                                    <div class="task-infos">
+                                        <p class="date"><i class="far fa-calendar"></i> 06/30/2020</p>
+                                        <p class="time"><i class="far fa-clock"></i> 4:00pm</p>
+                                    </div>
+                                </li>
+                            </ul>
+                        </div>
+                    </div>
+					<div id="chart-container">
+						<h2>Line Chart</h2>
```
peap, line 97: Could this title reflect what the chart will eventually show?

src/main/webapp/admin_dashboard.html
denzilbilson-google: These are meta tasks for the administrators to perform. I have changed the name of this div to "Administrator Tasks"

peap: Ah, thanks.

PR #15, opened 2020-06-29
404 page

denzilbilson-google:


src/main/webapp/404.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<link rel="stylesheet" href="404_style.css">
+	</head>
+	<body>
+		<div id="main">
+			<h1>Error 404</h1>
+			<p>Either you messed up or we did. Let's not play the blame game ; )</p>
```
LeonardZ0731, line 9: I like this error message 😆

src/main/webapp/404.html
```diff
@@ -0,0 +1,13 @@
...
+			<a href="/"><input type="button" id="button404" value="Go Back Home" /></a>
```
LeonardZ0731, line 10: Personally I would recommend using `<button onclick="window.location.href='/index.jsp';" id="button404">Go Back Home</button>`

src/main/webapp/404.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<link rel="stylesheet" href="404_style.css">
```
claudiatu, line 4: Could we add a page title here?

- [ ] add title

src/main/webapp/404_style.css
```diff
@@ -0,0 +1,41 @@
...
+p {
+	font-size: 2em;
+	text-align: center;
+	font-weight: 100;
+}
+
+h1 {
+	text-align: center;
+	font-size: 8em;
+	font-weight: 600;
```
claudiatu, line 24: It looks like we're loading in font weights 400, 100, 300 and 500 on line 1, but here we're using 600. Is this intended?

src/main/webapp/404.html
denzilbilson-google: Not familiar with jsp. Why would this be a better approach?

src/main/webapp/404.html
denzilbilson-google: Thanks!

src/main/webapp/404.html
LeonardZ0731: Ah it's actually not jsp, it's just plain html. It's just a button with a onclick function that redirects you to another url, which has the same functionality as your line 10. So there is no need to change.

peap:
No need for my review; now that claudiatu has write access, her approval counts toward the requirement of getting two.

PR #16, opened 2020-06-30
Implement the two-step verification feature for user profile page

LeonardZ0731:
Update the MyTaskServlet.java file to filter the tasks of status COMPLETE or COMPLETE: AWAIT VERIFICATION.
Update the user_profile.html file to add two tables that hold tasks of status COMPLETE & COMPLETE: AWAIT VERIFICATION
Update the script file to add functions that add tasks of status COMPLETE & COMPLETE: AWAIT VERIFICATION to the table. Also, implement verifyTask() and disapproveTask() that change a task of status "COMPLETE: AWAIT VERIFICATION" to status "COMPLETE" or "IN PROGRESS" respectively.


src/main/webapp/user_profile_script.js
```diff
@@ -160,8 +192,43 @@ async function displayNeedHelpTasks() {
     }
 }
 
+async function displayAwaitVerificationList() {
+    const queryURL = "/mytasks?keyword=Owner&complete=True";
+    const request = new Request(queryURL, {method: "GET"});
+    const response = await fetch(request);
+    const taskResponse = await response.json();
+    const awaitVerifBody = document.getElementById("await-verif-body");
```
poncem91, line 200: Same naming change as above :)

src/main/webapp/user_profile_script.js
```diff
@@ -80,22 +78,56 @@ async function abandonTask(keyString) {
     }
 }
 
+async function verifyTask(keyString) {
```
poncem91, line 81: I think it would be better to combine these two into an `updateTask(Keystring, status)` function, since they are very similar - the `completeTask` and `abandonTask` functions could be combined as well, but this is just my personal preference!

LeonardZ0731: Personally I don't think it's a good idea to combine the two functions because the alert messages are different. So even if I combine the two functions, I need extra if-else statements to send out different alert messages depending on the input `status`, which will also create a lot of code duplication. And for `completeTask` and `abandonTask`, their behavior is different. So I think I won't combine those two functions.

src/main/webapp/user_profile_script.js
```diff
@@ -160,8 +192,43 @@ async function displayNeedHelpTasks() {
     }
 }
 
+async function displayAwaitVerificationList() {
```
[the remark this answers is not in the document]
LeonardZ0731: Sure, that's a good idea. I forgot to change the name after including the complete tasks in the list. I would change the name to `displayNeedHelpCompleteTasks` in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure. I will modify this in the new commit.

src/main/java/com/google/sps/servlets/MyTaskServlet.java
```diff
@@ -43,6 +45,21 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     // on different fields accordingly.
     Filter filter = new FilterPredicate(keyword, FilterOperator.EQUAL, "Leonard");
 
+    // Depending on the input complete parameter, the status filter will be different.
+    String complete = request.getParameter("complete");
+    if (complete.equals("True")) {
+      Filter openFilter = new FilterPredicate("status", FilterOperator.NOT_EQUAL, "OPEN");
```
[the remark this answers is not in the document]
LeonardZ0731: Okay. I would modify this in the new commit.

src/main/webapp/user_profile_script.js
poncem91: Sounds good :) !

PR #17, opened 2020-06-30
update homepage with modal changes implemented in user_profile page

poncem91:


poncem91:
> You changed the font to Times New Roman, you might want to use the same font across pages
Just updated the fonts for consistency :) 

> It might be better to move the Event Listeners to your js files
Agreed! I do this in my users authentication branch, I'll merge that in after the prototype presentation.

src/main/webapp/homepage_style.css
```diff
@@ -165,47 +165,45 @@ section {
...
-#close-button:hover,
-#close-button:focus {
-  color: #000;
+
+.close-button:hover,
+.close-button:focus {
+  color: red;
   text-decoration: none;
   cursor: pointer;
 }
```
claudiatu, line 209: Could be worth adding a linter check to this repo for consistent formatting.
https://github.com/marketplace/actions/super-linter

poncem91: Thanks for this! I'll read through how to set this up for the repo.

PR #18, opened 2020-07-01
add feature that lists tasks in the homepage

poncem91:
Upon page load, the user's location is retrieved, converted to a neighborhood, and used to fetch and display tasks within the same neighborhood. Tasks are fetched by the TaskServlet's doGet function by filtering tasks that are within the same neighborhood (zipcode, country) and by category (if any has been specified). The servlet then constructs an HTML string with all the tasks that were returned by the prepared query and depending on whether the user is logged in and is the owner of said task the HTML will be presented differently. The response is sent back to the javascript file which then uses the response to display the tasks. Similarly, when choosing any of the categories, the same fetch request is made.


src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -29,9 +37,75 @@
...
+    query.setFilter(new Query.CompositeFilter(Query.CompositeFilterOperator.AND, filters));
+
+    // limits results to the 20 most recent tasks
+    List<Entity> results = datastore.prepare(query).asList(FetchOptions.Builder.withLimit(20));
+
+    StringBuilder out = new StringBuilder();
+
+    // Builds and stores HTML for each task
+    for (Entity entity : results) {
+      out.append("<div class='task'>");
```
poncem91, line 74: I tried using the requestDispatcher to send this over to a jsp template file instead, as it seemed like the better approach. Upon page load the jsp file would successfully receive and print (for testing) the tasks sent over to the terminal, however these results were shortlived (in the terminal it would print out the array of tasks and shortly after print out an empty array) and I was also not able to find a method that allows loading the jsp into the task-list div container after the template has been generated.
Any advice on how to better approach this is more than welcomed :)

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -29,9 +37,75 @@
...
+  // doGet method retrieves tasks from datastore and responds with the HTML for each task fetched
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Serve the GET request sent from home page to fetch all the tasks
-    return;
+    UserService userService = UserServiceFactory.getUserService();
+    boolean userLoggedIn = userService.isUserLoggedIn();
+    String userId = userLoggedIn ? userService.getCurrentUser().getUserId() : "null";
+
+    String zipcode = request.getParameter("zipcode");
+    String country = request.getParameter("country");
```
LeonardZ0731, line 47: Maybe you can add an error checking statement here for the case that zipcode and country is empty or invalid. request.getParameter() can return null if it fails to get the parameter specified and at that time you can send an error.

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -29,9 +37,75 @@
...
+    if (request.getParameterMap().containsKey("category")) {
+      String category = request.getParameter("category");
+      filters.add(new Query.FilterPredicate("category", Query.FilterOperator.EQUAL, category));
+    }
+
+    // Applies filters to query
+    query.setFilter(new Query.CompositeFilter(Query.CompositeFilterOperator.AND, filters));
+
+    // limits results to the 20 most recent tasks
+    List<Entity> results = datastore.prepare(query).asList(FetchOptions.Builder.withLimit(20));
```
LeonardZ0731, line 68: If we limit results to the 20 most recent tasks, I think we'd better notify the users on the web page. So, I would recommend adding a piece of information saying that "We limit to 20 most recent tasks" to `index.jsp`.

src/main/java/com/google/sps/servlets/TaskServlet.java
```diff
@@ -62,15 +136,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    taskEntity.setProperty("userId", userId);
     taskEntity.setProperty("detail", taskDetail);
     taskEntity.setProperty("timestamp", creationTime);
     taskEntity.setProperty("reward", rewardPts);
     taskEntity.setProperty("status", "OPEN");
     taskEntity.setProperty("Owner", "Leonard");
     taskEntity.setProperty("Helper", "N/A");
     taskEntity.setProperty("Address", "4xxx Cxxxxx Avenue, Pittsburgh, PA 15xxx");
+    taskEntity.setProperty("zipcode", "59715");
+    taskEntity.setProperty("country", "United States");
```
LeonardZ0731, line 153: Will we be able to decide the zipcode and country automatically with Maps API? If not, or if it is really hard, I can add two more fields that let user input their country and zipcode.

src/main/java/com/google/sps/servlets/TaskServlet.java
poncem91: That's smart, I'll implement some error checking in here, thanks!

src/main/java/com/google/sps/servlets/TaskServlet.java
poncem91: Good idea! I'll update this now.

src/main/java/com/google/sps/servlets/TaskServlet.java
poncem91: I don't think it should be too difficult, but that's not a bad idea to start with - that way we can stop using hardcoded data and once we get around it we can change it so that it retrieves the zipcode and country from the address. That would be great, thanks!

PR #32, opened 2020-07-06
implement claim tasks functionality

### LINKED ISSUE #22: Feature that lets helpers claim tasks from the homepage

### LINKED ISSUE #47: Race condition when claiming tasks to help out with
A race condition was present when claiming tasks, if a users clicked to help out with a task while it was open and in the middle of the process another user did the same thing, with the task still being open, then the 2nd user would be the one assigned as a helper - which isn't our desired outcome.

poncem91:
Logged in users can now help out with open tasks by clicking on the 'HELP OUT' button in a task from the homepage task list. Upon confirming to help, a fetch POST request is sent to the EditTaskServlet that checks that this is in fact a 'helpout' edit action, then checks that the user is indeed logged in and that the task hasn't been already claimed by another user. If none of these errors occur, the task's status is changed to 'IN PROGRESS' and the helper's userId is updated in the task's 'helper' property.

Resolves #22 , fixes #47

### DIFF
--- a/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
@@ -86,14 +86,13 @@ public class EditTaskServlet extends HttpServlet {
       String input = request.getParameter("task-detail-input");
       // If the input is valid, set the taskDetail value to the input value
       if (input != null) {
-        taskDetail = input;
+        taskDetail = input.trim();
       }
 
       // If input task detail is empty, reject the request to edit and send a 400 error.
       if (taskDetail.equals("")) {
         System.err.println("The input task detail is empty");
-        response.sendError(
-            HttpServletResponse.SC_BAD_REQUEST, "The task detail field cannot be empty.");
+        response.sendRedirect("/400.html");
         return;
       }
 
@@ -105,7 +104,6 @@ public class EditTaskServlet extends HttpServlet {
 
       response.sendRedirect("/user_profile.html");
     }
-
   }
 
   // Both TaskServlet and EditTaskServlet use this method. I will fix this by putting the function
--- a/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
@@ -105,6 +105,7 @@ public class EditTaskServlet extends HttpServlet {
 
       response.sendRedirect("/user_profile.html");
     }
+
   }
 
   // Both TaskServlet and EditTaskServlet use this method. I will fix this by putting the function
--- a/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps.servlets;
+package com.google.neighborgood.servlets;
 
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
--- a/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
@@ -20,8 +20,10 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.datastore.Transaction;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
+import com.google.neighborgood.helper.RewardingPoints;
 import java.io.IOException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
@@ -39,14 +41,6 @@ public class EditTaskServlet extends HttpServlet {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Entity task;
 
-    try {
-      task = datastore.get(taskKey);
-    } catch (EntityNotFoundException e) {
-      System.err.println("Unable to find the entity based on the input key");
-      response.sendError(HttpServletResponse.SC_NOT_FOUND, "The requested task could not be found");
-      return;
-    }
-
     UserService userService = UserServiceFactory.getUserService();
     if (!userService.isUserLoggedIn()) {
       System.err.println("User must be logged in to edit a task");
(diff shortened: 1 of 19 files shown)


src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -20,44 +20,25 @@
...
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that handles the request for editing the details and rewards for a certain task. */
+/** Servlet that handles the request for editing tasks. */
 @WebServlet("/tasks/edit")
 public class EditTaskServlet extends HttpServlet {
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String keyString = request.getParameter("task-id");
-    int rewardPts = getRewardingPoints(request, "reward-input");
```
poncem91, line 35: Just a heads up: this hasn't been deleted, just moved to inside an else-block below

src/main/webapp/homepage_script.js
```diff
@@ -43,48 +44,15 @@ function addUIClickHandlers() {
     }
 }
 
-/* Function adds all the necessary tasks 'click' event listeners*/
```
poncem91, line 46: This was also not deleted, but moved further down the script file so it's easier to read (it's now closer to the function that calls the function).

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -66,12 +47,66 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+            HttpServletResponse.SC_CONFLICT, "Task has already been claimed by another helper");
+      }
+
+      String userId = userService.getCurrentUser().getUserId();
+      task.setProperty("Helper", userId);
+      task.setProperty("status", "IN PROGRESS");
+      datastore.put(task);
+
+      response.setContentType("text/html;");
+      response.getWriter().println();
```
LeonardZ0731, line 73: If there is no input in the `println()` function, I think you can just remove the two lines of code.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -66,12 +47,66 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // Edits tasks that have been claimed by setting the "helper" property to the userId
+    // of the helper and changing the task's status to "IN PROGRESS"
+    if (request.getParameter("action").equals("helpout")) {
+
+      UserService userService = UserServiceFactory.getUserService();
+      if (!userService.isUserLoggedIn()) {
+        System.err.println("User must be logged in to help out with a task");
+        response.sendError(
+            HttpServletResponse.SC_UNAUTHORIZED, "You must be logged in to help out with a task");
+      }
```
LeonardZ0731, line 59: I would suggest putting this block of code outside of the if-else statement because I added the login status check in a new commit, so at the end both if block and else block will need to check whether the user has logged into an account. As a result, check whether the user has logged into an account can be put at the very beginning of the `doPost()` method.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -20,44 +20,25 @@
...
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
 import java.io.IOException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that handles the request for editing the details and rewards for a certain task. */
+/** Servlet that handles the request for editing tasks. */
```
LeonardZ0731, line 31: In my PR#21, I made some changes to this file. Thus, you may need to merge my newly-added code with this file.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: Sounds good, I'll move this to outside of the if-else statement.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: I'll leave this as is for now, since this branch is branching off Mafe-sps-to-neighborgood which branches from maste, and PR#21 hasn't been merged to master yet. But once it does I'll pull the new changes from master and merge them into this branch as well :)

LeonardZ0731: Sounds good!

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: You're right, I'll just use a return statement in here instead, thanks!

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -66,12 +47,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // of the helper and changing the task's status to "IN PROGRESS"
+    if (request.getParameter("action").equals("helpout")) {
+
+      if (!task.getProperty("status").equals("OPEN")) {
+        System.err.println("Task must be open to be claimed by a helper");
+        response.sendError(
+            HttpServletResponse.SC_CONFLICT, "Task has already been claimed by another helper");
+      }
+
+      String userId = userService.getCurrentUser().getUserId();
```
dandrews-code, line 68: Is there a potential race condition here if another user claims the task in between when the task is queried on line 43 and this change committed on line 71?  Does that mean that the first person would actually be the loser and the second person would end up with the task?  (I don't know anything about how datastore works, so it may have some transaction semantics that protect against this problem that I'm unaware of.)

poncem91: Thanks for pointing this out! I hadn’t heard of race conditions before but it does seem like this is an issue in this situation. It appears that in Datastore you can use Transactions to prevent this, so I’ll implement these changes to handle this edge case on the next commit.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -66,12 +47,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      task.setProperty("Helper", userId);
+      task.setProperty("status", "IN PROGRESS");
+      datastore.put(task);
+
+      return;
+
+      // Edits task's details and reward points
+    } else {
```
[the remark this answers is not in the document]
poncem91: You’re right, I’ll change this to improve readibility!

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -66,12 +47,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      int rewardPts = getRewardingPoints(request, "reward-input");
+      if (rewardPts == -1) {
```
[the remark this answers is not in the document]
poncem91: This is actually part of Leonard’s implementation that I moved into the else block, but I’ll mention this to him and we can modify this!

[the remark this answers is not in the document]
poncem91: @LeonardZ0731 this is what I mentioned :)

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -20,27 +20,94 @@
...
+    // Edits tasks that have been claimed by setting the "helper" property to the userId
+    // of the helper and changing the task's status to "IN PROGRESS"
+    if (request.getParameterMap().containsKey("action")) {
+      if (request.getParameter("action").equals("helpout")) {
+
+        // Makes use of Transactions to prevent race condition
+        Transaction transaction = datastore.beginTransaction();
+
+        try {
+          task = datastore.get(taskKey);
```
dandrews-code, line 61: Should this use the pattern shown here?  https://cloud.google.com/datastore/docs/concepts/transactions#datastore-datastore-transactional-update-java   So here, this would be transaction.get(taskKey).  

Here's their sample code I'm looking at:

```
Entity task;
Transaction txn = datastore.newTransaction();
try {
  task = txn.get(taskKey);
  if (task == null) {
    task = Entity.newBuilder(taskKey).build();
    txn.put(task);
    txn.commit();
  }
} finally {
  if (txn.isActive()) {
    txn.rollback();
  }
}
```

Thoughts?

poncem91: I implemented it that way first as well, but then I realized that we're using the AppEngine Datastore API, which differs slightly. This is the documentation I followed: https://cloud.google.com/appengine/docs/standard/java/datastore/transactions
I think I followed the pattern in those examples, but let me know if I missed anything :)

dandrews-code: Ok, I don't really understand the part you mentioned about the AppEngine Datastore API, but I see that this code matches the pattern in the link you provided, so LGTM.  :-)

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -20,27 +20,94 @@
...
+
+        } catch (EntityNotFoundException e) {
+          System.err.println("Unable to find the entity based on the input key");
+          response.sendError(
+              HttpServletResponse.SC_NOT_FOUND, "The requested task could not be found");
+          return;
+
+        } finally {
+          if (transaction.isActive()) {
+            transaction.rollback();
```
dandrews-code, line 83: Would it be useful to throw an error if we get to this point so the user knows their request failed?

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -20,27 +20,94 @@
...
+      System.err.println("User must be logged in to edit a task");
+      response.sendError(
+          HttpServletResponse.SC_UNAUTHORIZED,
+          "You must be logged in to perform this action on a task");
+    }
+
+    // Edits tasks that have been claimed by setting the "helper" property to the userId
+    // of the helper and changing the task's status to "IN PROGRESS"
+    if (request.getParameterMap().containsKey("action")) {
+      if (request.getParameter("action").equals("helpout")) {
```
dandrews-code, line 55: Can we join these two if statements?

if (request.getParameterMap().containsKey("action") && request.getParameter("action").equals("helpout")) {
  ...
}

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: Ah my bad. We’ve been using AppEngine to run our web apps during our internship and there’s a specific API for that that includes the use of Datastore within AppEngine.
Sounds good!

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: Thanks for pointing this out, I’ll add this!

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: You’re right! I was afraid the second statement would be evaluated even if the first evaluated to false, but I just learnt the double ampersand follows short-circuiting Logic. I’ll change this, thanks for the suggestion!

poncem91:
> LGTM for the java code. Want to ask @peap to look at the JS since I don't have much experience with front-end code?

Thanks and will do!

src/main/webapp/homepage_script.js
```diff
@@ -113,21 +81,37 @@ if (JSON.stringify(neighborhood) != JSON.stringify([null, null])) {
...
 
-/* Function that hides the task after user confirms they want to help out */
-function removeTask(element) {
-    element.closest(".task").style.display = "none";
+/* Function sends a fetch request to the edit task servlet when the user
+offers to help out, edits the task's status and helper properties, and
+then reloads the task list */
+function confirmHelp(element) {
+    const task = element.closest(".task");
+    const url = "tasks/edit?task-id=" + task.dataset.key + "&action=helpout";
```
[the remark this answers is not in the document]
peap: Also, it's more typical for query parameters in POST requests to appear in the body of the request, rather than the URL. Either way is fine, just letting you know.

To put the data in the body, you'd do something like:
```
const data = {"task-id": task.dataset.key, "action": "helpout"}
const request = new Request(url, {
    method: "POST",
    body: JSON.stringify(data),
    headers: {
        'Content-Type': 'application/json'
    }});
```

Some more examples at https://attacomsian.com/blog/using-javascript-fetch-api-to-get-and-post-data#using-fetch-to-post-data

I don't think you need to change this, but I wanted to bring it up. When you start having lots of parameters, putting them all in the URL starts to get more clunky.

src/main/webapp/homepage_script.js
```diff
@@ -113,21 +81,37 @@ if (JSON.stringify(neighborhood) != JSON.stringify([null, null])) {
...
+    const request = new Request(url, {method: "POST"});
+    fetch(request).then((response) => {
+        // checks if another user has already claimed the task
+        if (response.status == 409) {
+            window.alert
+                ("We're sorry, but the task you're trying to help with has already been claimed by another user.");
+            window.location.href = '/';
+        }
+        // fetches tasks again if user's current neighborhood was successfully retrieved and stored
+        else if (JSON.stringify(neighborhood) != JSON.stringify([null, null])) {
```
peap, line 106: In addition to my other comment about this comparison, it might be worth creating a simple function called something like `userNeighborhoodIsKnown` that can do this comparison for you whenever you need it. That'll make it easier to update what it means for the neighborhood to be known, should that ever change.

src/main/webapp/homepage_script.js
```diff
@@ -43,48 +44,15 @@ function addUIClickHandlers() {
...
 function filterTasksBy(category) {
+    currentCategory = category;
 
-// only fetches tasks if user's neighborhood has been retrieved
-if (JSON.stringify(neighborhood) != JSON.stringify([null, null])) {
-    fetchTasks(category)
-        .then(response => displayTasks(response));
-}
+    // only fetches tasks if user's neighborhood has been retrieved
+    if (JSON.stringify(neighborhood) != JSON.stringify([null, null])) {
```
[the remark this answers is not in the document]
poncem91: That works better, I'll make this change, thanks!

src/main/webapp/homepage_script.js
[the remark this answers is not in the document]
poncem91: This works better, I will make both of these changes :)
Thanks!

src/main/webapp/homepage_script.js
poncem91: That's a good idea! I'll add a function to deal with this instead.

src/main/webapp/homepage_script.js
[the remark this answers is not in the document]
poncem91: For some reason when I try using this method the fetch fails and throws the following error: `POST https://8080-dfe2475d-f262-4cff-8707-9112c4571804.us-west1.cloudshell.dev/tasks/edit 500 (Internal Server Error)` - strange. I'll keep it as before for now but I'm not really sure why this wasn't working.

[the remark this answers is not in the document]
peap: I'd check your server logs to see what the Java code is reporting. Usually a 500 error means something went wrong server-side.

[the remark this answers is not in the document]
poncem91: It seems the parameters aren't being passed through in the request. I wrote a debugging if-statement that prints out a message if the request doesn't find parameters for "task-id" and it printed out the message. The 500 error was happening because of a null pointer exception, from the missing parameter. The network tab also doesn't show the parameters in the request =/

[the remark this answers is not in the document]
peap: Hmm, it's probably not worth dwelling on then. We could debug it together at some point if you like, but we can move on for now.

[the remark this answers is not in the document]
poncem91: Sounds good!

PR #34, opened 2020-07-06
change homepage admin view so admins can also see userpage icon

poncem91:


src/main/webapp/index.jsp
```diff
@@ -20,7 +20,7 @@
       <!--Site Header-->
       <header>
           <nav>
-              <div id="userpage-icon">
+              <div id="user-icon">
```
LeonardZ0731, line 23: nit: Maybe it's better to call the `div` element as `dashboard-icon-container`, since it contains icons instead of being the icon itself.

src/main/webapp/index.jsp
```diff
@@ -33,7 +33,10 @@
               <%
               if (userLoggedIn && userService.isUserAdmin()) { 
               %>
-                  <a href="admin_dashboard.html">
+                  <a href="user_profile.jsp" class="user-icon">
```
LeonardZ0731, line 36: nit: I think calling the class as `dashboard-icon` can be a better idea because all the icons lead to a dashboard, either a dashboard for users or a dashboard for admins.

src/main/webapp/index.jsp
poncem91: Agreed! That sounds better

src/main/webapp/index.jsp
poncem91: I'll make this change, thanks!

src/main/webapp/index.jsp
```diff
@@ -20,7 +20,7 @@
       <!--Site Header-->
       <header>
           <nav>
-              <div id="userpage-icon">
+              <div id="dashboard-icon-container">
               <%
               if (userLoggedIn && !userService.isUserAdmin()) {
```
peap, line 25: Can you nest these two if-statements so we don't have to repeat the HTML for the User Page link?

poncem91: Yes, I'll make this change right now, thanks!

PR #38, opened 2020-07-06
Denzil admin prototype

denzilbilson-google:


src/main/java/com/google/sps/servlets/AdminPage.java
```diff
@@ -0,0 +1,57 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    Query query = new Query("Task").addSort("timestamp", SortDirection.DESCENDING);
+
+    PreparedQuery results = datastore.prepare(query);
+
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(results.asList(FetchOptions.Builder.withLimit(10))));
```
LeonardZ0731, line 52: If the admins can only see the 10 most recent tasks, I would prefer to add a notice on the admin page, saying that we will only show 10 most recent tasks here.

src/main/webapp/admin_dashboard.html
```diff
@@ -1,111 +1,92 @@
...
+            <div id="chart-div"></div>
+          </div>
+        </div>
+        <div class="flex-grid">
+          <div id="map-container">
+            <h2>Map</h2>
+            <div id="map-div"></div>
+          </div>
+          <div id="card-container">
+            <h2>Newest User Tasks</h2>
```
LeonardZ0731, line 82: As I've mentioned in the `AdminPage.java` file, it's great to mention that we only show the Newest 10 Users Tasks here.

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,30 +1,158 @@
...
+
+load(`https://maps.googleapis.com/maps/api/js?key=${map_key}`); // Add maps API to html
+
+google.charts.load("current", { packages: ["line"] });
 
+google.charts.setOnLoadCallback(drawChart);
 window.addEventListener("resize", drawChart);
+window.addEventListener("load", drawMap);
+
+//window.onload(drawMap());
```
LeonardZ0731, line 13: I think the commented line is no longer needed, you could remove this line. Just make `drawMap()` function an onload function for the `admin-dashboard.html` file.

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,30 +1,158 @@
...
+    [6, 8.8, 13.6, 7.7],
+    [7, 7.6, 12.3, 9.6],
+    [8, 12.3, 29.2, 10.6],
+    [9, 16.9, 42.9, 14.8],
+    [10, 12.8, 30.9, 11.6],
+    [11, 5.3, 7.9, 4.7],
+    [12, 6.6, 8.4, 5.2],
+    [13, 4.8, 6.3, 3.6],
+    [14, 4.2, 6.2, 3.4],
+  ]);
```
LeonardZ0731, line 36: Personal suggestion: Make the input of the `addRows()` as a variable for now. For example, define `var rowData = [[1, 37.8, 80.8, 41.8], ..., [14, 4.2, 6.2, 3.4]];` and change line 21 to `data.addRows(rowData);`. After making the change, if we want to replace the hard-coded data with real data, we can just set `rowData` to real data and don't need to make any further changes.

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,30 +1,158 @@
...
+  var chart = new google.charts.Line(document.getElementById("chart-div"));
+  chart.draw(data);
+}
+
+function drawMap() {
+  // Styles a map in night mode.
+  var map = new google.maps.Map(document.getElementById("map-div"), {
+    center: { lat: 40.674, lng: -73.945 },
```
LeonardZ0731, line 44: If the `lat` and `lng` values are fixed, I would recommend declaring these values as `const` at the beginning of the script file. For example, `const XXXLAT = 40.674; const XXXLNG = -73.945;`.

src/main/java/com/google/sps/servlets/AdminPage.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+class Comment {
```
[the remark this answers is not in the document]
denzilbilson-google: It's a left over class from something I was trying out. Got rid of it.

src/main/java/com/google/sps/servlets/AdminPage.java
```diff
@@ -0,0 +1,49 @@
...
+    PreparedQuery results = datastore.prepare(query);
+
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(results.asList(FetchOptions.Builder.withLimit(10))));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {}
+}
```
peap, line 49: It's probably clearer to omit this doPost method entirely until we make it do something.

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,30 +1,161 @@
-google.charts.load('current', {'packages':['line']});
-google.charts.setOnLoadCallback(drawChart);
+var map;
+var styledMapType;
+var map_key = keys.MAP_KEY;
```
peap, line 3: The README for this repo says we need `config.MAPS_KEY` in our config.js file. Can we standardize on that or `keys.MAP_KEY` so that we don't have to duplicate it in the config file?

src/main/java/com/google/sps/servlets/AdminPage.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+@WebServlet("/user-tasks")
```
peap, line 31: Just double-checking: is "user-tasks" the right name for what this returns to the admin page? And can we reuse another servlet that already knows how to return Task data?

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,30 +1,161 @@
...
+
+function load(file) {
+  var src = document.createElement("script");
+  src.setAttribute("type", "text/javascript");
+  src.setAttribute("src", file);
+  document.getElementsByTagName("head")[0].appendChild(src);
 }
+
+function initUserTasks() {
+  fetch("/data")
```
peap, line 146: My browser gets a 404 for this URL. Should it be "/user-tasks"?

src/main/java/com/google/sps/servlets/AdminPage.java
denzilbilson-google: Yes, this does return user tasks. We have talked amongst ourselves about creating a single servlet to do this, but decided that this is  not the best time to do so. It will definitely be implemented later.

peap: OK, sounds good. Thanks!

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,8 +1,8 @@
 var map;
 var styledMapType;
-var map_key = keys.MAP_KEY;
+var mapKey = config.MAP_KEY;
```
peap, line 3: Note that the variable name in the README is `MAPS_KEY`, not `MAP_KEY`.

peap: Is there a commit to push that renames this variable to `MAPS_KEY`, or updates the README to `MAP_KEY`?

src/main/webapp/user_profile_script.js
```diff
@@ -34,19 +34,7 @@ async function deleteTask(keyString) {
...
-        window.alert("You can only edit an 'OPEN' task.")
-    } else {
-        document.getElementById("edit-detail-input").value = info.detail;
-        document.getElementById("edit-point-input").value = info.reward.toString();
-        const id_input = document.getElementById("task-id-input");
-        id_input.value = info.keyString;
-        document.getElementById("editTaskModalWrapper").style.display = "block";
-        showNeedHelp()
-    }
-}
```
peap, line 49: I'm not sure how this deletion fits in to the PR. Was it intentional? This function is still called on line 168 of this file.

src/main/webapp/admin_dashboard.html
```diff
@@ -88,5 +88,19 @@ <h2> Last 10 User Tasks</h2>
         </div>
       </main>
     </div>
+    <div id="popup-overlay">
+		<div class="popup">
+			<h2>Edit Info for Task</h2>
+			<form action="">
+				<input type="text" placeholder="Task Details" id="task-details">
+
+				<input type="text" placeholder="Category" id="category">
```
peap: For a future PR: it's still good to give these elements `name` attributes so that the form works semantically. Even if you're just getting their values in the JavaScript by referencing the IDs of the elements.

src/main/webapp/admin_dashboard.html
```diff
@@ -88,5 +88,19 @@ <h2> Last 10 User Tasks</h2>
...
+				<input type="submit" value="Submit">
+			</form>
+			<a href="#" class="popup-close">&times;</a>
+		</div>
+	</div>
+	
+	<h1>Click <a href="#popup-overlay">here</a> to open the popup box</h1>
```
peap, line 104: For a future PR: hyperlinks with just "here" as the inner text are confusing when read aloud to users by a screen reader; we should aim to make the text more descriptive of what the link does.

PR #39, opened 2020-07-07
implement scrolling homepage UI behavior

### LINKED ISSUE #35: Implement homepage UI scrolling behaviors

poncem91:
resolves #35

### DIFF
--- a/src/main/webapp/homepage_script.js
+++ b/src/main/webapp/homepage_script.js
@@ -15,6 +15,22 @@
 const MAPSKEY = config.MAPS_KEY
 let neighborhood = [null , null];
 
+window.onscroll = stickyControlBar;
+
+/* Scroll function so that the control bar sticks to the top of the page */
+function stickyControlBar() {
+    const controlBarWrapper = document.getElementById("control-bar-message-wrapper");
+    const taskListDiv = document.getElementById("tasks-list");
+    const offset = 165;
+    if (window.pageYOffset >= offset || document.body.scrollTop >= offset || document.documentElement.scrollTop >= offset) {
+        controlBarWrapper.style.position = "fixed";
+        taskListDiv.style.marginTop = "165px";
+    } else {
+        controlBarWrapper.style.position = "static";
+        taskListDiv.style.marginTop = "auto";
+    }
+}
+
 /* Calls addUIClickHandlers and getUserNeighborhood once page has loaded */
 if (document.readyState === 'loading') {
     // adds on load event listeners if document hasn't yet loaded
--- a/src/main/webapp/homepage_script.js
+++ b/src/main/webapp/homepage_script.js
@@ -19,8 +19,8 @@ window.onscroll = stickyControlBar;
 
 /* Scroll function so that the control bar sticks to the top of the page */
 function stickyControlBar() {
-    const controlBarWrapper = document.getElementById("control-bar-message-wrapper");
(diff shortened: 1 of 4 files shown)


src/main/webapp/homepage_script.js
```diff
@@ -15,6 +15,22 @@
 const MAPSKEY = config.MAPS_KEY
 let neighborhood = [null , null];
 
+window.onscroll = stickyControlBar;
+
+/* Scroll function so that the control bar sticks to the top of the page */
+function stickyControlBar() {
+    const controlBarWrapper = document.getElementById("control-bar-message-wrapper");
+    const taskListDiv = document.getElementById("tasks-list");
+    const offset = 165;
```
denzilbilson-google, line 24: Common javascript syntax would be to put constants in all caps to differentiate them from regular variables.

src/main/webapp/homepage_style.css
```diff
@@ -65,6 +69,17 @@ section {
   padding-bottom: 1.5em;
 }
 
+#control-bar-message-wrapper {
+  width: 100%;
+  max-width: 55em;
+  z-index: 11;
```
denzilbilson-google, line 75: It seems that this is the only z-index element. I may be mistaken, but I believe setting the value to 1 should have the same effect as a lone z-index with a value of 11

src/main/webapp/homepage_script.js
poncem91: You're right, I'll update this, thanks!

denzilbilson-google:
Are you planning to pass the linter checks before merging or will you merge them and deal with the linter later?

src/main/webapp/homepage_style.css
poncem91: Ah yes, I need to stop using random numbers for this, I'll change this to 1 since it's the only one there. Thanks!

poncem91:
> Are you planning to pass the linter checks before merging or will you merge them and deal with the linter later?

Yup to the first thing! I'll merge the master branch into this branch which should take care of the lint checks (the same files have been already modified in the master branch so merging that in should fix things). I want to keep the PR clean for now so it doesn't confuse reviewers, but once I get reviewer's approval I'll make the lint checks. By default the lint checker won't let you merge into master without passing the tests

src/main/webapp/homepage_script.js
```diff
@@ -21,8 +21,8 @@ window.onscroll = stickyControlBar;
 function stickyControlBar() {
     const controlBarWrapper = document.getElementById("control-bar-message-wrapper");
     const taskListDiv = document.getElementById("tasks-list");
```
denzilbilson-google: Sorry to nitpick, but these two are also const and it might be best to change them

poncem91: I'm going to have to disagree with this one, if that's alright. These are not primitive types but objects whose properties are going to change. Looking at some of the ToTT resources, they also use const without uppercase on some occasions, so I'd rather keep this in camelCase to avoid readers thinking that the HTML object itself is immutable when only its reference is.

denzilbilson-google: If their properties are going to change would it not be better to use "let" or "var" over "const"?

poncem91: I want to keep the variable reference constant so it always points to the same element and doesn't get reassigned to a different element. But either way works, I'll just change this to 'let' to simplify things!

PR #46, opened 2020-07-08
Google OSS setup

peap:
Add Apache license, contribution guidelines, and disclaimer in README


peap:
> Thank you for adding the licenses! LGTM. However, the style of the CSS files seem to have some problems. But I believe that Denzil has already fixed this in PR#38. After PR#38 get approved, I think a merge will be needed.

Makes sense, thanks! I'll wait for #38 to get merged.

PR #54, opened 2020-07-09
Add test cases for servlet files

LeonardZ0731:


src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,419 @@
...
+      dummy.setProperty("nickname", "Leonard" + str);
+      dummy.setProperty("address", "xxx" + str);
+      dummy.setProperty("phone", str + str);
+      dummy.setProperty("email", "test" + str + "@example.com");
+      dummy.setProperty("userId", str);
+      dummy.setProperty("points", 0);
+      ds.put(dummy);
+    }
+
+    // Check the number of entities in the datastore. The result should be 10,000
```
poncem91, line 373: nit: I think the comment might need to be updated to 999 :)

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,419 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** */
+@RunWith(JUnit4.class)
+public final class UserInfoServletTest {
+
+  private LocalServiceTestHelper helper =
```
poncem91, line 49: It might be helpful adding a comment here to explain what this does :)

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Ah sure. I forgot to update the comments. I will fix it in the new commit.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Definitely. I added some comments to illustrate the setup of the testing environment.

peap:
Generated files are normally not committed to the repo; could you add `/WEB-INF/appengine-generated` to `.gitignore` and remove the local_db.bin file from the repo?

LeonardZ0731:
Sure. I will remove it now.

src/test/java/com/google/neighborgood/TaskServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+    response = mock(HttpServletResponse.class);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void setUpTest() {
```
peap, line 79: This name is similar enough to `setUp` that it made me wonder if it was part of the JUnit framework; can we call it something like `helpersTest` or `testEnvironmentTest`?

...five minutes later I see that it has a `@Test` annotation, but it was still confusing to me :)

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+      assertTrue(false);
+    }
+
+    // After sending the GET request, the doGet function should output the json string
+    writer.flush();
+    assertTrue(stringWriter.toString().contains("[\"Leonard\",\"xxx\",\"xxx\"]"));
+  }
+
+  @Test
+  public void stressTest() {
```
peap, line 357: This one fails for me sometimes (it's flaky). More about that below.

I see this error:

```
-------------------------------------------------------------------------------
Test set: com.google.neighborgood.servlets.UserInfoServletTest
-------------------------------------------------------------------------------
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.122 sec <<< FAILURE! - in com.google.neighborgood.servlets.UserInfoServletTest
stressTest(com.google.neighborgood.servlets.UserInfoServletTest)  Time elapsed: 3.457 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.neighborgood.servlets.UserInfoServletTest.stressTest(UserInfoServletTest.java:407)
```

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+            new Query("UserInfo")
+                .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, "1234567890")));
+    Entity entity = results.asSingleEntity();
+
+    Instant end = Instant.now();
+    Duration timeElapsed = Duration.between(start, end);
+
+    // Fetching the specific user info from datastore should take a reasonable amount of time
+    // I think 10ms is a reasonable threshold value.
+    assertTrue(timeElapsed.toMillis() < 10);
```
peap, line 407: While it's good to track the performance of our site, we should really attempt to make our tests stable. Furthermore, this is testing the performance of a local version of a dependency, which isn't really under our control. We can trust the Datastore helper to be reasonably performant in our unit tests.

This also brings up a question for me - have you set up indexes on the columns (like userId) that you intend to query frequently? If you do that, you can expect Datastore to perform adequately in these lookups, regardless of the amount of data.

If you do want to add an end-to-end test with junit, we could tag it and create a different Maven target that includes tests with the tag you choose. Then it won't run this test every time, which will be nice while developing the site, since those kinds of tests can be flaky and/or time consuming. See https://www.baeldung.com/junit-filtering-tests, for example.

src/test/java/com/google/neighborgood/TaskServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    // After sending the POST request, there should be one entity in the datastore
+    assertEquals(1, ds.prepare(new Query("Task")).countEntities(withLimit(10)));
+    PreparedQuery results = ds.prepare(new Query("Task"));
+    Entity entity = results.asSingleEntity();
+    if (entity == null) {
+      // The entity can't be null
+      assertTrue(false);
+    }
```
peap, line 112: `assertNotNull(entity)` is a more concise way to express this

src/test/java/com/google/neighborgood/TaskServletTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    // Test the stored task information
+    assertEquals("1234567890", (String) entity.getProperty("Owner"));
+    assertEquals("Help me please", (String) entity.getProperty("detail"));
+    assertEquals(50, (long) entity.getProperty("reward"));
+  }
+
+  @Test
+  public void multipleInputDoPostTest() {
```
peap, line 121: Can the three `doPost` calls be condensed into a for-loop?

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+
+    when(request.getParameter("nickname-input")).thenReturn("Leonard");
+    when(request.getParameter("address-input")).thenReturn("4xxx Centre Avenue");
+    when(request.getParameter("phone-input")).thenReturn("4xxxxxxxxx");
+
+    try {
+      new UserInfoServlet().doPost(request, response);
+    } catch (IOException e) {
+      assertTrue(false);
+    }
```
peap, line 105: In your tests, I think any exception raised will automatically stop the test and cause it to fail, so we don't need to try-catch these `doPost` calls. We can let the test framework take care of failing the test when there's an exception we're not expecting.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+    assertEquals("Leo", (String) entity.getProperty("nickname"));
+    assertEquals("4xxx Centre Avenue", (String) entity.getProperty("address"));
+    assertEquals("4xxxxxxxxx", (String) entity.getProperty("phone"));
+    assertEquals("leonardzhang@google.com", (String) entity.getProperty("email"));
+    assertEquals("1234567890", (String) entity.getProperty("userId"));
+    assertEquals(0, (long) entity.getProperty("points"));
+  }
+
+  @Test
+  public void edgeCaseTest() {
```
peap, line 236: Can the method name be more descriptive about what edge case(s) you're testing here? If it's multiple types of edge cases, it's better to give them all their own test method.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+    when(request.getParameter("phone-input")).thenReturn("4xxxxxxxxx");
+
+    try {
+      new UserInfoServlet().doPost(request, response);
+    } catch (IOException e) {
+      assertTrue(false);
+    }
+
+    // The POST request should lead to an error handling clause, there should be no entity in the
+    // datastore
```
peap, line 256: It would be good to assert something about the response, as well, such as the status code that is send back to the browser when the input is bad.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+
+    try {
+      new UserInfoServlet().doGet(request, response);
+    } catch (IOException e) {
+      assertTrue(false);
+    }
+
+    // After sending the GET request, the doGet function should output the json string
+    writer.flush();
+    assertTrue(stringWriter.toString().contains("[\"Leonard\",\"xxx\",\"xxx\"]"));
```
peap, line 353: It would be good to test the content type header of the response object, as well, to ensure that the server tells the client that it is returning JSON.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
...
+    when(request.getParameter("address-input")).thenReturn("4xxx Centre Avenue");
+    when(request.getParameter("phone-input")).thenReturn("4xxxxxxxxx");
+
+    try {
+      new UserInfoServlet().doPost(request, response);
+    } catch (IOException e) {
+      assertTrue(false);
+    }
+
+    // After sending the POST request, there should be three entities in the datastore
```
peap, line 390: not three, but one thousand

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -0,0 +1,422 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
LeonardZ0731: The -Xlint:deprecation tag doesn't work on the cloud shell terminal, but I manged to figure out that the deprecation warning is caused by the stress test. I will try to fix it in the new commit.

src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps.helper;
```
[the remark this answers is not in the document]
LeonardZ0731: I think these two files have already been fixed. I just checked this branch and didn't find any instance of "sps" in the two files.

src/test/java/com/google/neighborgood/TaskServletTest.java
LeonardZ0731: Ah I think `testEnvironmentTest` is definitely a better name for the test case. I will change its name in the new commit. The current name is confusing 😭.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: This flaky behavior is caused by the stressTest, due to the running time of a search is not stable. As discussed yesterday, I decided to delete the stress test for now, and consider to add it back after I find a better, more stable way to carry out stress test.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Ah I actually have not set up indexes on the columns on the userId. But it's really a great idea. After setting it up, most queries should be handled really efficiently. I will record this as an issue and implemented it after the MVP.

And as discussed in chat yesterday, I decided to remove the stress tests for now. But I will look at the JUnit filtering test later on to implement a stable stress test suite for the servlet files.

src/test/java/com/google/neighborgood/TaskServletTest.java
LeonardZ0731: Ah this is way more concise. I will switch to this in the new commit.

src/test/java/com/google/neighborgood/TaskServletTest.java
LeonardZ0731: Sure. I will try to use a single for-loop for the three `doPost` calls and test the result in the loop.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Okay. I will remove all these try-catch clauses and declared that the functions may throw IOException in the new commit.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: It is mainly a test for the edge case that the input field is empty or just spaces. I change the function name to `emptyInputTest()` in the new commit and add a comment to illustrate its purpose.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: For my current implementation, the error handling clause will print an error message and redirect the user to a 400 error page. So I think I would test the system output here since it's hard to test the page redirect.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Sure. I tried multiple ways to get the content type header of the response object, but since it's a mock object, the content type header is always null. So the only thing I can do is to use Mockito to verify that the servlet file has tried to set its content type to `application/json`.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
LeonardZ0731: Ah I forgot to change the comments. The stress test is removed in the new commit so the problem has been fixed automatically.

src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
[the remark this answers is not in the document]
peap: Those ones are gone now (after a merge?), but now I see it in `src/main/java/com/google/sps/servlets/AdminPage.java`.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
peap: Sounds good. Can you also assert something about the status code itself, e.g. using `response.getStatus()`?

src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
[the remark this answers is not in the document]
LeonardZ0731: I think Denzil has already fixed them in his PR.

PR #55, opened 2020-07-09
implement top scores board

### LINKED ISSUE #27: Implement user scoreboard that showcases the top scorers

poncem91:
Resolves #27 

Users can click on the top right link in the homepage to open up the top scorers board. Open opening the two tables fetch data from the UserInfoServlet that gathers the top 10 scorers. Data for world top scorers is gathered and loaded into the table, and then if the user's current neighborhood is known, it also gathers and loads the top 10 scorers in the neighborhood near the user. Additionally, if the user makes it to the scoreboard, their name shows up in bold and includes a congratulatory HTML title attribute.

Note: the user's zipcode and country stored in datastore is currently hardcoded, so if you wish to test the scoreboard that shows neighbor top scorers you'll want to change the zipcode and country to yours (or this has also proved useful for me for faking different current locations: https://developers.google.com/web/tools/chrome-devtools/device-mode/geolocation )

### DIFF
--- a/src/main/java/com/google/neighborgood/User.java
+++ b/src/main/java/com/google/neighborgood/User.java
@@ -22,7 +22,7 @@ public final class User {
   private final String phone;
   private final String email;
   private final String userId;
-  private final int points;
+  private final long points;
 
   public User(Entity entity) {
 
@@ -31,6 +31,6 @@ public final class User {
     this.phone = (String) entity.getProperty("phone");
     this.email = (String) entity.getProperty("email");
     this.userId = (String) entity.getProperty("userId");
-    this.points = (Integer) entity.getProperty("points");
+    this.points = (Long) entity.getProperty("points");
   }
 }
--- a/src/main/java/com/google/neighborgood/User.java
+++ b/src/main/java/com/google/neighborgood/User.java
@@ -25,6 +25,7 @@ public final class User {
   private final String email;
   private final String userId;
   private final long points;
+  private boolean isCurrentUser;
 
   public User(Entity entity) {
 
@@ -36,5 +37,14 @@ public final class User {
     this.email = (String) entity.getProperty("email");
     this.userId = (String) entity.getProperty("userId");
     this.points = (Long) entity.getProperty("points");
+    this.isCurrentUser = false;
+  }
+
+  public void isCurrentUser() {
+    this.isCurrentUser = true;
+  }
+
+  public String getUserId() {
+    return this.userId;
   }
 }
--- a/src/main/java/com/google/neighborgood/User.java
+++ b/src/main/java/com/google/neighborgood/User.java
@@ -1,0 +1,36 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.neighborgood;
+
+import com.google.appengine.api.datastore.Entity;
+
+public final class User {
+  private final String nickname;
+  private final String address;
+  private final String phone;
+  private final String email;
+  private final String userId;
+  private final int points;
+
+  public User(Entity entity) {
+
+    this.nickname = (String) entity.getProperty("nickname");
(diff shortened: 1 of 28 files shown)


src/main/java/com/google/neighborgood/User.java
```diff
@@ -0,0 +1,50 @@
...
+    this.zipcode = (String) entity.getProperty("zipcode");
+    this.country = (String) entity.getProperty("country");
+    this.phone = (String) entity.getProperty("phone");
+    this.email = (String) entity.getProperty("email");
+    this.userId = (String) entity.getProperty("userId");
+    this.points = (Long) entity.getProperty("points");
+    this.isCurrentUser = false;
+  }
+
+  public void isCurrentUser() {
```
LeonardZ0731, line 43: nit: The function name `isCurrentUser()` makes me feel like this function will return a boolean, determining whether the user is the current user. So I think maybe change the name to something sounds like a verb can be a good idea. I think `setAsCurrentUser()` can be a good option.

src/main/java/com/google/neighborgood/servlets/UserInfoServlet.java
```diff
@@ -94,6 +148,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       entity = new Entity("UserInfo");
       entity.setProperty("nickname", nickname);
       entity.setProperty("address", address);
+      entity.setProperty("zipcode", "59715");
+      entity.setProperty("country", "United States");
```
LeonardZ0731, line 152: I made some changes to the doGet() function in the branch leonard-map. Maybe it will cause a confilct with your file.

poncem91: I'll check the branch out to see what conficts we can avoid :) 
thanks for the heads up!

src/main/java/com/google/neighborgood/User.java
poncem91: That's a much better name for the method, thanks for the suggestion, I'll change this!

src/main/java/com/google/neighborgood/servlets/UserInfoServlet.java
poncem91: Ah I just saw what you meant, I'll keep an eye out when merging the branch but will leave those hardcoded for now since we can't test my branch with the real zipcode, country values yet :)

src/main/java/com/google/neighborgood/servlets/UserInfoServlet.java
```diff
@@ -38,14 +39,53 @@
 public class UserInfoServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     UserService userService = UserServiceFactory.getUserService();
 
+    // Retrieves user accounts for the topscorers board
+    if (request.getParameterMap().containsKey("action")
```
dandrews-code, line 47: Am I reading this correctly that users don't need to be logged in to view the top 10 list?  Is that what we want?

src/main/java/com/google/neighborgood/servlets/UserInfoServlet.java
```diff
@@ -38,14 +39,53 @@
...
+        && request.getParameter("action").equals("topscorers")) {
```
dandrews-code, line 48: Any thoughts on breaking out the contents of this if {...} to a separate function just for keeping this function small and easier for readers to understand what is going on in it?

poncem91: That's a good idea, thanks! I will create a private function that executes the code inside the If-block.

src/main/java/com/google/neighborgood/servlets/UserInfoServlet.java
poncem91: I had actually thought of allowing guest users to see the top10 lists as well - do you think this is a bad idea? I can change this to be only visible to logged users if that's the case.

dandrews-code: I like the idea!  My only thought is about how this performs a query/sort of the entire UserInfo table which sounds like it could slow down the server if lots of random users load this page.  I think a typical pattern for this problem is to have a background task on the server that periodically performs this computation and then writes the output to a "top 10" table in the database, and then external users loading the "top 10" page will read from this "top 10" table.  I have no idea how easy/hard this is to do with this server framework, so I'll let you decide if this is reasonable or not.  :-)

poncem91: This is actually a really good idea, thanks! I'm not 100% certain how to implement this either, but I'll look into it after the MVP presentation and add it to our list of 'issues' so I can hopefully address this in an upcoming PR :)

PR #57, opened 2020-07-09
modify TaskServlet to retrieve user's nickname and show in task

### LINKED ISSUE #56: tasks in homepage should show the user's nickname and not id

poncem91:
Resolves #56

### DIFF
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -112,7 +112,7 @@ public class TaskServlet extends HttpServlet {
         return;
       }
 
-      out.append("<div class='username'>")
+      out.append("<div class='user-nickname'>")
           .append((String) userEntity.getProperty("nickname"))
           .append("</div>");
       if (userLoggedIn) {
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -20,7 +20,10 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.datastore.Query.SortDirection;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
@@ -93,8 +96,17 @@ public class TaskServlet extends HttpServlet {
       }
       out.append("<div class='task-container'>");
       out.append("<div class='task-header'>");
+
+      // Retrieves task's user to get user's nickname
+      String taskUserId = (String) entity.getProperty("userId");
+      Query userQuery =
+          new Query("UserInfo")
+              .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, taskUserId));
+      PreparedQuery userResults = datastore.prepare(userQuery);
+      Entity userEntity = userResults.asSingleEntity();
+
       out.append("<div class='username'>")
-          .append((String) entity.getProperty("Owner"))
+          .append((String) userEntity.getProperty("nickname"))
           .append("</div>");
       if (userLoggedIn) {
         // changes the Help Button div if the current user is the owner of the task
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -184,7 +184,7 @@ public class TaskServlet extends HttpServlet {
 
     String userId = userService.getCurrentUser().getUserId();
 
-    // Retrieves current user entity to include as the task's parent
+    // Creates current user entity key to include as the task's parent
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Key userEntityKey = KeyFactory.createKey("UserInfo", userId);
 
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -115,7 +115,7 @@ public class TaskServlet extends HttpServlet {
           out.append(userNickname);
         } catch (EntityNotFoundException e) {
           System.err.println(
-              "Unable to find the user entity based on the current user id. Setting a default nickname.");
+              "Unable to find the task's owner info to retrieve the owner's nickname. Setting a default nickname.");
           out.append("Neighbor");
         }
       }
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -105,16 +105,15 @@ public class TaskServlet extends HttpServlet {
       PreparedQuery userResults = datastore.prepare(userQuery);
       Entity userEntity = userResults.asSingleEntity();
 
-      if (userEntity == null) {
-        System.err.println("Unable to find the user entity based on the current user id");
-        response.sendError(
-            HttpServletResponse.SC_NOT_FOUND, "The requested user could not be found");
-        return;
-      }
+      out.append("<div class='user-nickname'>");
 
-      out.append("<div class='user-nickname'>")
-          .append((String) userEntity.getProperty("nickname"))
-          .append("</div>");
+      if (userEntity == null) {
+        System.err.println(
+            "Unable to find the user entity based on the current user id. Setting a default nickname.");
+        out.append("Neighbor");
+      } else out.append((String) userEntity.getProperty("nickname"));
+
+      out.append("</div>");
       if (userLoggedIn) {
         // changes the Help Button div if the current user is the owner of the task
         if (!userId.equals((String) entity.getProperty("userId"))) {
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -31,6 +31,7 @@ import com.google.gson.Gson;
 import com.google.neighborgood.helper.RewardingPoints;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
@@ -83,6 +84,10 @@ public class TaskServlet extends HttpServlet {
 
     StringBuilder out = new StringBuilder();
 
+    // Stores task owner's user info to prevent querying multiple
+    // times in datastore for the same user's info
+    HashMap<String, String> usersNicknames = new HashMap<String, String>();
+
     // Builds and stores HTML for each task
     for (Entity entity : results) {
       out.append("<div class='task' data-key='")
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -93,8 +96,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       out.append("<div class='task-header'>");
+
+      // Retrieves task's user to get user's nickname
+      String taskUserId = (String) entity.getProperty("userId");
+      Query userQuery =
+          new Query("UserInfo")
+              .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, taskUserId));
+      PreparedQuery userResults = datastore.prepare(userQuery);
+      Entity userEntity = userResults.asSingleEntity();
+
```
LeonardZ0731, line 107: nit: I would recommend to add an error handling after line 106, checking whether `userEntity` is `null`. If it's null, we should return a 404 error message. You could refer to the `doPost()` function in `TaskInfoServlet.java` file.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -93,8 +96,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       }
       out.append("<div class='task-container'>");
       out.append("<div class='task-header'>");
+
+      // Retrieves task's user to get user's nickname
+      String taskUserId = (String) entity.getProperty("userId");
+      Query userQuery =
+          new Query("UserInfo")
+              .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, taskUserId));
```
LeonardZ0731, line 104: One thing: Is the `userId` field overlapped with the `Owner` field for the `UserInfo` entity? I don't remember that I've set a userId field before, but now I found that in the `doPost()` of `TaskServlet.java` file, a `userId` field is set for the `UserInfo` entity, but its content is the same as `Owner` field.

poncem91: Ah yes, I think I added at some point when the tasks were hard coded still. At that point Owner was the name/nickname, so they weren’t duplicated - we should definitely remove one of them. I’ll update this in the next commit (I think we’ve been using the “owner” one the most so far, so I’ll remove the userId one).

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: Good idea! I’ll add this in the next commit, thanks!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -93,8 +96,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      // Retrieves task's user to get user's nickname
+      String taskOwner = (String) entity.getProperty("Owner");
+      Query userQuery =
+          new Query("UserInfo")
+              .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, taskOwner));
+      PreparedQuery userResults = datastore.prepare(userQuery);
+      Entity userEntity = userResults.asSingleEntity();
+
+      if (userEntity == null) {
+        System.err.println("Unable to find the user entity based on the current user id");
```
peap, line 109: "current" is ambiguous; could you say something about how the task's owner couldn't be found?

Also, is it worth failing the whole request if we can't look up one user?

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -93,8 +96,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       out.append("<div class='task-container'>");
       out.append("<div class='task-header'>");
+
+      // Retrieves task's user to get user's nickname
+      String taskOwner = (String) entity.getProperty("Owner");
+      Query userQuery =
+          new Query("UserInfo")
+              .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, taskOwner));
+      PreparedQuery userResults = datastore.prepare(userQuery);
+      Entity userEntity = userResults.asSingleEntity();
```
peap, line 106: Thinking about the context of this code, what would you expect to happen when there are lots of tasks in play?

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -93,8 +96,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+      if (userEntity == null) {
+        System.err.println("Unable to find the user entity based on the current user id");
+        response.sendError(
+            HttpServletResponse.SC_NOT_FOUND, "The requested user could not be found");
+        return;
+      }
+
       out.append("<div class='username'>")
```
peap, line 115: Might be worth changing this class name in we're showing nicknames now, instead of usernames. Once you merge your CSS rules, it's possible you'll want to display usernames and nicknames differently, in which case this could get displayed in an inconsistent way.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: Perhaps it would be better to have a fallback user nickname for when the actual nickname could not be found (eg. "Neighbor"). I'll make both of these changes for this, thanks for bringing this up!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: Since only 20 tasks are being queried every time the homepage loads and the stress test done by Leonard on retrieving users's info returned results in less or about 10ms, the extra time that it takes to retrieve each task's user's nickname should remain around 0.2s, which I think is reasonable. However, the task query for gathering the most recent tasks might be more time-consuming as there are more and more tasks at play, as it must sort all the tasks before knowing which are the 20 most recent. So  the response time will grow with the task size (I wasn't able to find documentation on what sorting algorithm datastore uses but if it is O(nlogn) as many sorting algorithms are, then response time for sorting tasks would grow linearithmiclly with the total number of tasks). As we aren't expecting many thousands (millions?) ) of tasks (not to mention the datastore limitation of 1000 entitities), I think this is reasonable as well, but let me know if you think this is a faulty assumption!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: This is a good idea, I'll change this to nicknames, thanks! Should we merge the CSS files into one for our project or keep each page with its own CSS file? I looked up online what the typical approach was but found conflicting answers so I wasn't sure if we should merge them or no at the end.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
peap: Just curious, where does the limit of 1000 entities come from? I'm not seeing it here: https://cloud.google.com/datastore/docs/concepts/limits.

Having just reviewed #54, I think I can say that the 10ms number comes from using a local testing version of Datastore, rather than what you'll be working with while running in production. Now, it might be better or worse than that in production, but I don't it's necessarily been measured. We can also trust the database to have a reasonable sorting algorithm, so we don't need to be too worried about using the Datastore API to do the ordering for us.

Anyway, for now you're only fetching 20 tasks at a time, sure. Still, you're making an additional round-trip to Datastore for each of those tasks to look up its owner's nickname (which you might have already looked up for previous tasks). I generally try to avoid doing database lookups in a loop, preferring to get more information out of (perhaps more complicated) queries. Can you think of a way to reduce the number of database lookups in this handler?

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
peap: I suspect you'll have some similarities across pages that you want to share, so one option would be to have a `base.css` file and then keep page-specific ones. Then have every page use the base CSS alongside its own file. That can of course happen in another PR.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
peap: sounds good to me

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: Leonard came across that 1000 entities limitation when searching for the testing documentation, I think he meant this part: "Maximum number of keys allowed for a Lookup operation in the Datastore API". I think this means that if we have more than 1000 entities we can't do `datastore.get(taskKey)`, but we'd still be able to do queries, so it might not be as much of a limitation as we thought.

Ah, that makes sense about the 10ms being just the local testing version. 

To avoid as many queries, specially for tasks from the same owner, I could cache the user nicknames, and check if it's been retrieved already before doing another database lookup. I think that would increase performance, specially if there are multiple tasks by the same user. I could also use projection queries to get just the user's nickname from the UserInfo entities, this would also reduce lookup cost. I'm not sure I got what you meant with avoiding database lookups in a loop and instead getting more info out of queries. I'm not sure it's possible to use one query across multiple entity kinds unless there is an ancestor relationship, but I don't think that applies here since the ancestor in this case would be the user.

Thanks for pushing me to find all these alternative solutions to enhancing the request's performance, I really appreciate it!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: That makes sense, thanks for clarifying!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -160,25 +184,34 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(taskEntity);
+    Key userEntityKey = KeyFactory.createKey("UserInfo", userId);
 
-    // Redirect back to the user page.
-    response.sendRedirect("/user_profile.jsp");
+    try {
+      Entity userEntity = datastore.get(userEntityKey);
+      // Create an Entity that stores the input comment
+      Entity taskEntity = new Entity("Task", userEntity.getKey());
```
poncem91, line 194: I added the task's parent when creating the new Entity.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -160,25 +184,34 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    taskEntity.setProperty("category", "misc");
-
+    // Creates current user entity key to include as the task's parent
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(taskEntity);
+    Key userEntityKey = KeyFactory.createKey("UserInfo", userId);
 
-    // Redirect back to the user page.
-    response.sendRedirect("/user_profile.jsp");
+    try {
```
poncem91, line 191: The below lines of code are mostly the same, I just moved everything into a try-catch statement due to the addition of line 192.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
peap: > I think he meant this part: "Maximum number of keys allowed for a Lookup operation in the Datastore API". I think this means that if we have more than 1000 entities we can't do `datastore.get(taskKey)`,

I interpret this as `datastore.get(taskKey1, taskKey2, taskKey3, ...)` only allows you to look up 1000 tasks at a time *by their keys*. I think you can have more entities than that, but a single Lookup operation (done by `datastore.get`) can only ask for 1000 by their keys explicitly.

And yep, good ideas about reducing the number of database queries. I like the solution you've come up with where you only look them up when needed. In the future, I might consider doing a single project query to get all the user nicknames and put them into a map, but of course that starts to get out of control when you have a lot of users.

poncem91: That makes a lot of sense, it’s good to know we can exceed 1000 entities!

Sounds good! :)

As for the suggestion of doing a project query and storing it in a map - would we need to use something like an http session to store the map so it doesn’t go away after the request is done or how are these things usually handled?

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: My bad, I will change this so it only includes the one line. Thanks!

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
peap: You're on the right track - you might eventually set up some kind of external cache that persists across sessions and is available to all replicas of the server. Or, you could store the result with a short timeout ("memoization") in some variable with a broader scope than the doGet function, so that future requests on that server don't have to re-do the lookup for some amount of time. (The user session is typically used for this-user-specific data (e.g., shopping cart items, user preferences, etc.), rather than something like this.)

But even request-specific caching of user lookups is an improvement over doing N user queries for every HTTP request to this handler. I don't see a need to go beyond that in this PR.

poncem91: This makes perfect sense, thanks for clarifying! I might look into this further after going over some of the post-MVP features. It'd be nice to cache these externally if time allows.

PR #66, opened 2020-07-10
Add a map to the User Info Modal that allows the user to input their personal address, zipcode and country by clicking on the map

LeonardZ0731:


src/main/webapp/account.jsp
```diff
@@ -4,6 +4,8 @@
     <meta charset="UTF-8">
     <title>My Personal Info</title>
     <link rel="stylesheet" href="account_style.css">
+    <script src="https://maps.googleapis.com/maps/api/js?key=YOUR_API_KEY_HERE&libraries=places&language=en;"></script>
```
poncem91, line 7: Just for consistency, I would recommend using the config.js file to store the API key, like we did for the Admin and Homepage, that way all of our files are using the same method to retrieve the API key.

LeonardZ0731: I tried to use the config.js file to store the API key but the Geocoding API always fails if I didn't include the key in the jsp file, which is really weird.

poncem91: That's odd, did you include this before your maps API script tag? `<script type='text/javascript' src='config.js'></script>`

LeonardZ0731: I did, but the Geocoding services didn't give me any response.

poncem91: Ah, I think including a JS variable doesn't work in HTML/JSP, so you'd have to dynamically load the script tag within your initMap() function. I'd try adding the following at the beginning of your initMap() function:

const script = document.createElement("script");
script.type = "text/javascript";
script.src =  "https://maps.googleapis.com/maps/api/js?key=" + MAPSKEY + "&callback=init&libraries=places&language=en";
script.defer = true;
script.async = true;
document.head.appendChild(script);

window.init = async function() {
   --your current initMap() function would go inside here--
}

LeonardZ0731: Cool! I would try it out tomorrow. 👍

peap: +1 - let's get all the maps working using the same key-loading mechanism

src/main/webapp/user_profile_script.js
```diff
@@ -301,4 +308,277 @@ async function displayOfferHelpCompleteTasks() {
...
+function displayMarker(position) {
+    var lat = position.lat();
+    var lng = position.lng();
+    let marker = new google.maps.Marker({
+        position: {lat: lat, lng: lng},
+        map: map
+    });
+
+    google.maps.event.addListener(marker, "dblclick", function(event) {
+        deleteMarker(lat, lng);
```
peap, line 532: Should we also clear the address and zip code inputs when deleting a marker?

src/main/webapp/account.jsp
LeonardZ0731: I just made a new commit and switched to the same key-loading mechanism as Mafe and Denzil.

src/main/webapp/account.jsp
```diff
@@ -12,7 +14,7 @@
   if (!userService.isUserLoggedIn()) {
        response.sendRedirect(userService.createLoginURL("/account.jsp"));
   } else if (RetrieveUserInfo.getInfo(userService) == null) { %>
-  <body>
+  <body onload="initMap()">
```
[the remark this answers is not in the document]
LeonardZ0731: After switching to the new key-loading mechanism, this has been moved to js files and is no longer in jsp files. So yes, this is not set as an onload attribute in the jsp file now.

src/main/webapp/account.jsp
```diff
@@ -33,7 +35,24 @@
                     <br/>
                 </div>
                 <br/>
-                <textarea name="address-input" id="address-input" required="true" placeholder="Input your address here:"></textarea>
+                <textarea name="address-input" id="edit-address-input" required="true" placeholder="Input your address here:"></textarea>
+                <p id="rest-map">Click to mark your personal address on the map!</p>
+                <input id="place-input" class="controls" type="text" placeholder="Search Box">
+                <div id="map"></div>
+                <br/><br/>
```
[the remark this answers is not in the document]
LeonardZ0731: Ah definitely we can adjust the padding in the CSS file, but there will be a lot of paddings to be changed. So I would create an issue for this problem and addressed it in another PR later on.

src/main/webapp/user_profile.jsp
```diff
@@ -24,7 +26,7 @@
         String logoutURL = userService.createLogoutURL(urlToRedirectAfterUserLogsOut);
         String nickname = userInfo.get(0);
         String points = userInfo.get(3); %>
-  <body onload="showNeedHelp()">
+  <body onload="initialize()">
```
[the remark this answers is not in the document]
LeonardZ0731: The account.jsp is a page that allows the users to input their personal information. If a user haven't input his personal information, he will not be allowed to access the user_profile.jsp page and he will be redirected to account.jsp page automatically.

src/main/webapp/user_profile_script.js
```diff
@@ -301,4 +308,277 @@ async function displayOfferHelpCompleteTasks() {
...
+    map = new google.maps.Map(document.getElementById("map"), {
+        center: {lat: GOOGLE_KIRKLAND_LAT, lng: GOOGLE_KIRKLAND_LNG},
+        zoom: 18,
+        styles: [
+            {
+                "elementType": "geometry",
+                "stylers": [{"color": "#ebe3cd"}]
+            },
+            {
+                "elementType": "labels.text.fill",
```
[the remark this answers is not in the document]
LeonardZ0731: I think the style of the maps on my `account.jsp` and `user_profile.jsp` page is different from the style of the map on Denzil's admin page. But we can definitely centralize the map creation after this PR. I will also record this as an issue.

src/main/webapp/account.jsp
```diff
@@ -33,7 +35,24 @@
...
+                <div>
+                    <label for="edit-zipcode-input">Your Zip Code:</label>
+                    <br/>
+                </div>
+                <br/>
+                <textarea name="zipcode-input" id="edit-zipcode-input" required="true" placeholder="Input your zipcode here:"></textarea>
```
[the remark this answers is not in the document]
LeonardZ0731: Ah that's really a good point. A textarea box is too large for zipcode, country and phone number input. So I change all of the three inputs to text-type input in the new commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure. I should definitely do that. In the new commit, when the user deletes a marker, the address, zipcode and country input will be set to empty and a placeholder would be shown to remind the user to fill out the blank.

src/main/webapp/user_profile.jsp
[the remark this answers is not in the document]
peap: Ah, ok. To me, a "user profile" page and an "account" page are basically the same thing., so I have a renaming suggestion: maybe `account.jsp` could become something like `account_setup.jsp` to indicate to developers more clearly what that page is for.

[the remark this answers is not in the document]
LeonardZ0731: Ah I got the point. Yeah, the account page sounds like a place for user to view their personal account info, so maybe a name such as "personal_info_setup.jsp" or "account_setup.jsp" is a better name. I would create a new PR for that.

src/main/webapp/user_profile_script.js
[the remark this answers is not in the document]
peap: Sounds good. I see you filed #77 for this, and I would consider unifying the map styles between the profile and admin pages as part of that issue, as well.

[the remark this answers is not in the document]
LeonardZ0731: Sure. I would add that to issue 77.

PR #70, opened 2020-07-13
Denzil admin prototype

denzilbilson-google:


src/main/webapp/admin_dashboard.html
```diff
@@ -1,122 +1,126 @@
...
+						<h2> Last 10 User Tasks</h2>
+						<div id="card-div">
+							<ul id="user-tasks">
+							</ul>
+						</div>
+					</div>
+				</div>
+			</main>
+		</div>
+		<div id="popup-overlay">
```
LeonardZ0731, line 107: Is it possible to change the style of the edit task popup window? The color scheme can stay the same but I think it would be better if the edit task modal has the same size as other pages' pop up window

src/main/webapp/admin_dashboard_script.js
```diff
@@ -29,166 +29,176 @@ window.addEventListener("load", getUserTasks());
...
+				taskSection.innerHTML += addTask(userTask);
+			}
+		});
 }
 
 function addTask(task) {
-  let string = `<a href="#popup-overlay"><li class="admin-task user-task"  onclick="openWithPopup(${task.key.id})"><h3> ${task.propertyMap.category} </h3>`;
-  string += `<h4> ${task.propertyMap.Owner} </h4></li></a>`;
-  return string;
+	let string = `<a href="#popup-overlay"><li class="user-task"  onclick="openWithPopup('${task.keyString}')"><h3> ${task.category} </h3>`;
```
LeonardZ0731, line 175: Same suggestion here: Maybe "admin-task" is a better name than "user-task" for the class

src/main/webapp/admin_dashboard_style.css
```diff
@@ -182,7 +182,7 @@ main h1 {
   opacity: 1;
 }
 
-.popup {
+#popup {
```
LeonardZ0731, line 185: As I've mentioned above, it would be great if you adjust the size of the popup window to make it the same as the edit task pop up window on user profile page

src/main/java/com/google/neighborgood/servlets/AdminPage.java
```diff
@@ -0,0 +1,125 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+@WebServlet("/user-tasks")
```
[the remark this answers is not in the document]
denzilbilson-google: These are the user tasks for the admin page, I have renamed it to admin-user-tasks

src/main/webapp/admin_dashboard_script.js
denzilbilson-google: renamed to admin-user-task

src/main/webapp/admin_dashboard_style.css
denzilbilson-google: I believe we discussed compatability and uniformity would be a post-MVP consideration.

src/main/webapp/admin_dashboard.html
denzilbilson-google: The styles across pages will be dealt with after the MVP

src/main/java/com/google/neighborgood/servlets/AdminPage.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.neighborgood.task.Task;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns some example content. TODO: modify this file to handle comments data */
```
poncem91, line 37: nit: might be best to remove/edit this comment

PR #80, opened 2020-07-15
add fallback IP location to getUserLocation function

poncem91:


src/main/webapp/homepage_script.js
```diff
@@ -221,11 +221,15 @@ function getUserNeighborhood() {
...
-            }, function(err) {
+        navigator.geolocation.getCurrentPosition(function(position) {
+            var location = {lat: position.coords.latitude, lng: position.coords.longitude};
+            resolve(location);
+        }, function(err) {
+            let url = "https://www.googleapis.com/geolocation/v1/geolocate?key=" + MAPSKEY;
+            const request = new Request(url, {method: "POST"});
+            fetch(request).then(response => response.json()).then(jsonresponse => {
+                resolve(jsonresponse["location"]);
+            }).catch(() => {
```
peap, line 232: I think this breaks the assumptions of the `LOCAL_DEV_LAT_LNG` error handler. Now it belongs to an error handler for a fetch request, rather than `navigator.geolocation.getCurrentPosition`, so the error codes it might receive are probably different. Also, the geolocation fetch request seems less likely to fail (unless the API key is bad), so the local dev fallback is less likely to be used.

Can we keep the `LOCAL_DEV_LAT_LNG` override associated with the initial `navigator.geolocation.getCurrentPosition` call?

src/main/webapp/homepage_script.js
```diff
@@ -221,11 +221,15 @@ function getUserNeighborhood() {
 /* Function that returns a promise to get and return the user's location */
 function getUserLocation() {
     return new Promise((resolve, reject) => {
-        if (navigator.geolocation) {
-            navigator.geolocation.getCurrentPosition(function(position) {
-                var location = {lat: position.coords.latitude, lng: position.coords.longitude};
-                resolve(location);
-            }, function(err) {
+        navigator.geolocation.getCurrentPosition(function(position) {
```
peap, line 224: Do we know that `navigator.geolocation` always exists?

poncem91: My bad, I should've kept that if-statement. I wrongly thought `function(err)` would be able to handle those cases. I'll update this in the next commit.

src/main/webapp/homepage_script.js
poncem91: I see what you're saying about the fetch error codes not being the same, it seems the fetch request for the geolocation API returns errors 400-404 if something goes wrong with retrieving the location response. I would rather attempt to retrieve a fallback position using an IP address before using `LOCAL_DEV_LAT_LNG`. Would it seem reasonable to check the response of the fetch request for errors in the mentioned range, and if any of those errors were present then use `LOCAL_DEV_LAT_LNG`?

peap: Yep, that sounds fine to me. Everyone has an IP address (and hopefully a valid API key), so I think it makes LOCAL_DEV_LAT_LNG practically obsolete. That's fine, but I'd consider removing it at this point, either in this PR or another one.

poncem91: Sounds good! I'll remove it now, thanks!

PR #82, opened 2020-07-15
fix scrolling overlapping that was occurring in smaller screens

poncem91:


src/main/webapp/homepage_script.js
```diff
@@ -22,13 +22,16 @@ window.onscroll = stickyControlBar;
...
-    const OFFSET = 165;
-    if (window.pageYOffset >= OFFSET || document.body.scrollTop >= OFFSET || document.documentElement.scrollTop >= OFFSET) {
-        controlBarWrapper.style.position = "fixed";
-        taskListDiv.style.marginTop = "165px";
-    } else {
-        controlBarWrapper.style.position = "static";
-        taskListDiv.style.marginTop = "auto";
+
+    if (window.innerWidth > 1203) {
+        const OFFSET = 190;
```
LeonardZ0731, line 27: I would recommend to add some comments for this part. For me, 1203 and 190 are just magic numbers and I don't know the meaning of these two numbers. So it would be better to explain why you set the `OFFSET` to 190 and why the control bar should behaves differently when the `window.innerWidth` is larger than 1203

poncem91: Good idea, thanks for the suggestion! I will make these changes now.

PR #83, opened 2020-07-15
Add a new feature that a info window will pop up if the user clicks the task

LeonardZ0731:
The pop up info window will show the detail of the task

### DIFF
--- a/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
@@ -124,6 +124,22 @@ public class EditTaskServlet extends HttpServlet {
       return;
     }
 
+    // Get the task overview from the form input
+    String taskOverview = "";
+
+    input = request.getParameter("task-overview-input");
+    // If the input is valid, set the taskOverview value to the input value
+    if (input != null) {
+      taskOverview = input.trim();
+    }
+
+    // If input task overview is empty, reject the request to edit and send a 400 error.
+    if (taskOverview.equals("")) {
+      System.err.println("The input task overview is empty");
+      response.sendRedirect("/400.html");
+      return;
+    }
+
     // Get task category from the form input
     String taskCategory = request.getParameter("edit-category-input");
     if (taskCategory == null || taskCategory.isEmpty()) {
@@ -142,6 +158,7 @@ public class EditTaskServlet extends HttpServlet {
 
     // Set the details, category,and rewards to the newly input value
     task.setProperty("detail", taskDetail);
+    task.setProperty("overview", taskOverview);
     task.setProperty("reward", rewardPts);
     task.setProperty("category", taskCategory);
     datastore.put(task);
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -130,8 +130,11 @@ public class TaskServlet extends HttpServlet {
         }
       }
       out.append("</div>");
-      out.append("<div class='task-content'>")
-          .append((String) entity.getProperty("detail"))
+      out.append(
+              "<div class='task-content' onclick='showTaskInfo(\""
+                  + KeyFactory.keyToString(entity.getKey())
+                  + "\")'>")
+          .append((String) entity.getProperty("overview"))
           .append("</div>");
       out.append("<div class='task-footer'><div class='task-category'>#")
           .append((String) entity.getProperty("category"))
@@ -188,6 +191,21 @@ public class TaskServlet extends HttpServlet {
       return;
     }
 
+    // Get the task overview from the form input
+    String taskOverview = "";
+    input = request.getParameter("task-overview-input");
+    // If the input is valid, set the taskOverview value to the input value
+    if (input != null) {
+      taskOverview = input.trim();
+    }
+
+    // If input task overview is empty, reject the request to add a new task and send a 400 error.
+    if (taskOverview.equals("")) {
+      System.err.println("The input task overview is empty");
+      response.sendRedirect("/400.html");
+      return;
+    }
+
     long creationTime = System.currentTimeMillis();
 
     String userId = userService.getCurrentUser().getUserId();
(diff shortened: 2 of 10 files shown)


src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -124,6 +124,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    input = request.getParameter("task-overview-input");
+    // If the input is valid, set the taskOverview value to the input value
+    if (input != null) {
+      taskOverview = input.trim();
+    }
+
+    // If input task overview is empty, reject the request to edit and send a 400 error.
+    if (taskOverview.equals("")) {
+      System.err.println("The input task overview is empty");
+      response.sendRedirect("/400.html");
```
swparr, line 139: let's open an issue to track removal of the 400 redirects for these errors

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -188,6 +191,21 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    input = request.getParameter("task-overview-input");
+    // If the input is valid, set the taskOverview value to the input value
+    if (input != null) {
+      taskOverview = input.trim();
+    }
+
+    // If input task overview is empty, reject the request to add a new task and send a 400 error.
+    if (taskOverview.equals("")) {
+      System.err.println("The input task overview is empty");
+      response.sendRedirect("/400.html");
```
swparr, line 205: As as above

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
LeonardZ0731: Sure. I will open an issue right now.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
LeonardZ0731: Okay, I've created issue #84 for the 400 redirects removal.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
```diff
@@ -124,6 +124,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       return;
     }
 
+    // Get the task overview from the form input
```
poncem91, line 127: Personally task overview sounds too much like task detail, would task title reflect better what we're trying to achieve with this field?

src/main/webapp/homepage_style.css
```diff
@@ -313,3 +313,16 @@ section {
...
+
+#task-detail-container {
+  margin-bottom: 20px;
+}
+
+.task-content {
+  cursor: pointer;
+}
+
+.task-content:hover {
```
poncem91, line 325: To keep the page consistent, I'd rather you didn't use this styling but instead the same styling used by a:hover so it's consistent with the rest of the homepage.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -130,8 +130,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         }
       }
       out.append("</div>");
-      out.append("<div class='task-content'>")
-          .append((String) entity.getProperty("detail"))
+      out.append(
```
poncem91, line 133: I would prefer to have this onclick function dynamically added through the script file. Another suggestion would be to have the whole task div be clickable (sort of like the user tasks in the admin page). I'd like to work on this latter modification since I was planning on doing same changes to the homepage task list anyway, so I can go ahead and change the onclick part as well if you like.

src/main/webapp/index.jsp
```diff
@@ -131,5 +138,12 @@
             </form>
         </div>
     </div>
+    <div class="modalWrapper" id="taskInfoModalWrapper">
+        <div class="modal" id="taskInfoModal">
+            <span class="close-button" id="task-info-close-button" onclick="closeTaskInfoModal()">&times;</span>
```
poncem91, line 143: Same comment as above about the onclick function in here.

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
LeonardZ0731: I think overview is totally different from detail. Overview is just a general review or summary of a certain task, such as "Help me buy some food from a nearby supermarket", but task detail will include the shopping list and requirements. So I think overview is a good name choice here.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
LeonardZ0731: Sure. The user_profile page is kind of different from the main page, so it's definitely better if you could help me change the onclick part.

src/main/webapp/index.jsp
LeonardZ0731: Sure. Can you help me change this onclick function? Thanks.

src/main/webapp/homepage_style.css
LeonardZ0731: Sure. I will change it in the new commit

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
poncem91: Sounds good then! Would it be possible to make this distinction more explicit in the add task form then? Perhaps adds something that suggests the overview is meant to be one sentence only. having the input fields differ in size (or have a max set of characters allowed for the overview) might also help emphasize this distinction.

src/main/java/com/google/neighborgood/servlets/TaskServlet.java
poncem91: Sounds good!

src/main/webapp/index.jsp
poncem91: Yup! I'll make these changes, just opened a new issue for it #86

src/main/java/com/google/neighborgood/servlets/EditTaskServlet.java
LeonardZ0731: Ah I would try to add this feature in a separate PR. It's a little bit hard to make such kind of distinction obvious and I will try to figure out how to fix this

LeonardZ0731: I have created an issue #87 for this problem

PR #85, opened 2020-07-16
implement integration testing with webdriver(chrome)

poncem91:
I'm still missing tests for the scoreboard and edge case testing, but these many lines of code seemed already way too much for a single PR :)

In order to run the IntegrationTest you must have Google Chrome installed and it must be accessible from where you are running the IntegrationTest. I'd recommend [running this from the linux terminal with Chrome for linux installed](https://docs.google.com/document/d/1ZbKjd08XL-Q_F6Jxe5A0feC6R59d55H2MuQ0FyZvmMI/edit). If the test is overall running too slow, try closing out of all your chrome tabs and also possibly restarting your computer. The webdriver occasionally will freeze loading a page making the test flaky - if a page is loading for over 30 seconds you might have to refresh the page or restart the test altogether. I've implemented some functionality to try to prevent these situations as much as possible but it still happens occasionally. Overall, if everything is running well, the test shouldn't take more than 30-120 seconds.

**Note:** Change the value of the constant USER_ZIPCODE (line 57) to equal a string of your current zipcode.
 
To run the IntegrationTest first run `mvn clean`, then start the devServer with the command `mvn package appengine:start` and once the devServer is running use the command `mvn -Dtest=IntegrationTest test` (this test class has been removed from the tests that run with `mvn test` due to its potential to take long periods of time to load and its potential to be flaky). **Important**: Once you are done testing, stop the devServer with command `mvn package appengine:stop` - if you forget to do this you'll run into issues running other devServers again.


src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+        wait.until(
+                new Function<WebDriver, WebElement>() {
+                  public WebElement apply(WebDriver driver) {
+                    return driver.findElement(By.id("no-tasks-message"));
+                  }
+                })
+            .isDisplayed();
+    // Message alerting user there are no tasks nearby should be displayed since there are no tasks
+    // yet in the site
+    assertTrue("No tasks in neighborhood message should be displayed", taskResultsMessageDisplayed);
```
LeonardZ0731, line 135: Is there a way to check whether the correct message is shown on the page? Instead of checking whether there is an element of ID "no-tasks-message", it would be better to check whether the element contains the correct info "Sorry, there are currently no tasks within your neighborhood for you to help with."

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    // Confirm that the user is still logged in
+    verifyLoggedUser(USER_NICKNAME, "login-logout");
+
+    // Randomizes task contents
+    Random random = new Random();
+    String taskDetail = TASK_DETAIL + random.nextInt(1000);
+    String taskOverview = TASK_OVERVIEW + random.nextInt(1000);
+    String rewardPoints = Integer.toString(random.nextInt(201));
+    int categoryOptionIndex = random.nextInt(TASK_CATEGORIES.length);
+    String taskCategory = TASK_CATEGORIES[categoryOptionIndex];
```
LeonardZ0731, line 162: I would suggest adding multiple tasks instead of only one task here. Maybe adding a loop to add multiple tasks can be a good idea

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+
+    addTask(taskDetail, rewardPoints, categoryOptionIndex, taskOverview);
+
+    // User should be redirected to user profile page after adding a task
+    assertTrue(
+        "User should be in user profile page",
+        driver.getCurrentUrl().contains("/user_profile.jsp"));
+
```
LeonardZ0731, line 170: I'm not familiar with the tools you used here, but I wonder if there is a way to check how many `<tr>` elements are present on the new user page. We can count the number of `<tr>` elements and make sure that the number of `<tr>` elements is equal to the number of tasks that we have added.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    verifyNewTaskHomepage();
+  }
+
+  @Test
+  /** Tests functionality of adding tasks from the userpage */
+  public void _03_UserPage_AsLoggedUser_AddTask() {
+    goToUserPage();
+    ifLaggingThenRefresh();
+
+    // Randomizes task contents
```
LeonardZ0731, line 184: As the suggestions above, adding multiple tasks and make sure there are correct amount of `<tr>` elements present on the page can be a great idea.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+
+    addTask(taskDetail, rewardPoints, categoryOptionIndex, taskOverview);
+
+    // User should be redirected to user profile page after adding a task
+    assertTrue(
+        "User should be in user profile page",
+        driver.getCurrentUrl().contains("/user_profile.jsp"));
+
+    driver.manage().timeouts().implicitlyWait(5, TimeUnit.SECONDS);
+    verifyNewTaskUserPage();
```
LeonardZ0731, line 172: I think you could add a really brief comment here, explaining the functionality of function `verifyNewTaskUserPage()` and `verifyNewTaskHomepage()`. It would help the readers understand why we need to have these functions here. While reviewing this, I need to scroll the page up and down to understand code from line 172 to 175, so it would be really helpful if you could explain the purpose of these functions here with comments.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    driver.manage().timeouts().implicitlyWait(5, TimeUnit.SECONDS);
+
+    // Refreshes page due to flakiness of test caused by the partial refresh of the div in the page.
+    // Without refreshing this often resulted in a a stale element error
+    driver.navigate().refresh();
+    goToOfferHelp();
+
+    // Location of first task listed in the complete task table in userpage
+    String taskCompletedXPath = "//tbody[@id='complete-task-body']/tr[1]";
+
```
LeonardZ0731, line 258: Can we use a for loop here instead of having so many duplicated code? For example, how about creating an array of String such as `String[] testTargets = new String[]{"overview", "status", "nickname" ... etc.};` Then create a for loop, during the loop we set a value of a String variable `temp` to the result of `wait.until` and make an assertion on the variable `temp` during each loop. The change of `/td[i]` can also be achieved by using the index.

Using a loop can save a lot of spaces and make the code cleaner. 😄

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+  public void _07_Userpage_AsLoggedUser_VerifyCompletedTask() {
+    // Logs out by using 'logout-href' element id;
+    logOut("logout-href");
+    loginUser(USER_EMAIL);
+    ifLaggingThenRefresh();
+
+    // Location of first task listed in the await verification table
+    String awaitVerifTaskXPath = "//tbody[@id='await-verif-body']/tr[1]";
+
+    String taskOverview =
```
LeonardZ0731, line 313: As the suggestion above for function `_06_Userpage_AsLoggedHelper_CompleteTask()`, using a foor loop here can be great.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+        .click();
+
+    // Clicking on verify button triggers an alert confirmation window
+    // Driver will try to accept the alert to verify the task every second for a minute
+    for (int i = 0; i < 60; i++) {
+      try {
+        driver.switchTo().alert().accept();
+        break;
+      } catch (NoAlertPresentException e) {
+        driver.manage().timeouts().implicitlyWait(1, TimeUnit.SECONDS);
```
LeonardZ0731, line 361: I'm a little bit confused about all these `implicitlyWait` functions here. Are they really necessary? If they are, can we reduce the time we need to wait for?

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    // Helper's total points
+    String points =
+        wait.until(
+                new Function<WebDriver, WebElement>() {
+                  public WebElement apply(WebDriver driver) {
+                    return driver.findElement(By.id("points"));
+                  }
+                })
+            .getText();
+    assertEquals("My current points: " + Integer.toString(helperPoints) + "pts", points);
```
LeonardZ0731, line 406: I'm confused about how the value of helperPoints gets set here. Is this test cases definitely executed after the former testcase gets executed?

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+        driver.switchTo().alert().accept();
+        break;
+      } catch (NoAlertPresentException e) {
+        driver.manage().timeouts().implicitlyWait(1, TimeUnit.SECONDS);
+      }
+    }
+    // updates recentTask
+    recentTask.put("status", "IN PROGRESS");
+
+    String taskStatus =
```
LeonardZ0731, line 463: Same suggestion here: Using a loop to replace these three blocks of code can be a good idea

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    for (int j = 1; j < allEntityKinds.size(); j++) {
+      // clicks on list button
+      wait.until(
+              new Function<WebDriver, WebElement>() {
+                public WebElement apply(WebDriver driver) {
+                  return driver.findElement(By.id("list_button"));
+                }
+              })
+          .click();
+      driver.manage().timeouts().implicitlyWait(5, TimeUnit.SECONDS);
```
LeonardZ0731, line 561: We have already waited for the previous code block getting executed, why we still have to wait for 5 seconds here?

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+  }
+
+  /**
+   * Verifies that newly added task's details are correctly display in userpage's need help table
+   */
+  private void verifyNewTaskUserPage() {
+    // Location of most recent task in user page's need help
+    String taskRowXPath = "//table[@id='need-help']/tbody/tr[1]";
+
+    String rowTaskOverview =
```
LeonardZ0731, line 651: Same suggestion here: Using for loop can be a great idea.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+    verifyTaskDetails(taskRowXPath + "/td[1]");
+  }
+
+  /** Verifies that newly added tasks are displayed properply in homepage */
+  private void verifyNewTaskHomepage() {
+
+    // First task location in homepage
+    String taskXPath = "//div[@id='tasks-list']/div[1]/div[2]";
+
+    String taskOverviewActual =
```
LeonardZ0731, line 690: Still we can try to use a for loop to replace code from line 690 to 718. For example, `for (int i = 1; i < 4; i++) ` and when i equals to 2, we can treat it differently.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+
+  /**
+   * Verifies that after claiming a task/offering help with a task, that it displays correctly in
+   * the user page's offer help table
+   */
+  private void verifyOfferHelpTask() {
+    // Location of most recent task in offer help table in userpage
+    String offerHelpRowXPath = "//tbody[@id='offer-help-body']/tr[1]";
+
+    String taskOverview =
```
LeonardZ0731, line 900: Same suggestion here: A loop can be a better choice

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1065 @@
...
+   * Updates the stored instance of recentTask with the most recent open task's contents displayed
+   * in the homepage
+   */
+  private void updateRecentTask() {
+    recentTask.clear();
+
+    // Most recent task location in homepage
+    String taskXPath = "//div[@id='tasks-list']/div[1]/div[2]";
+
+    String taskNickname =
```
LeonardZ0731, line 957: Same suggestion as the suggestion for `verifyNewTaskHomepage()`: using a loop here and treat the situation when i == 2 differently

src/main/webapp/homepage_script.js
```diff
@@ -45,9 +45,9 @@ if (document.readyState === 'loading') {
 
 /* Function adds all the necessary UI 'click' event listeners*/
 function addUIClickHandlers() {
-    // adds showCreateTaskModal and closeCreateTaskModal click events for the add task button
-    if (document.body.contains(document.getElementById("addtaskbutton"))) {
-        document.getElementById("addtaskbutton").addEventListener("click", showCreateTaskModal);
+    // adds showModal and closeModal click events for the add task button
```
LeonardZ0731, line 48: Since the event listeners are `showCreateTaskModal` and `closeCreateTaskModal`, should the comments be "adds 'showCreateTaskModal' and 'closeCreateTaskModal' click events for the add task button" instead?

src/main/webapp/homepage_script.js
```diff
@@ -58,7 +58,7 @@ function addUIClickHandlers() {
             filterTasksBy(e.target.id);
         });
     }
-    // adds showTopScoresModal click event 
+    // adds showTopScoresModal click event
```
LeonardZ0731, line 61: It will be great to mention `closeTopScoresModal` click event here.

src/main/webapp/user_profile.jsp
```diff
@@ -42,7 +42,7 @@
         <div id="container">
             <button class="help-button" id="need-help-button" onclick="showNeedHelp()">Need help</button>
             <button class="help-button" id="offer-help-button" onclick="showOfferHelp()">Offer help</button>
-            <i class="fa fa-plus-circle fa-3x" id="create" onclick="showModal()"></i>
+            <i class="fa fa-plus-circle fa-3x" id="create-task-button" onclick="showModal()"></i>
```
LeonardZ0731, line 45: Is the name of the onclick function incorrect? I think the function has been renamed to `showCreateTaskModal` right?

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: There's a different div for every message (the other ones are not displayed - set to display:none) so by checking that that one div is displayed we are checking that the correct message is displayed :)

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: That's a good idea, I'll do that, thanks!

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: It sounds like there should be a way to do this, I'll look into this and add this too - thanks for the suggestion!

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: Sounds good! I'll add some comments in here.

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: This is a great suggestion, thanks! I'll make these changes now

src/test/java/com/google/neighborgood/IntegrationTest.java
LeonardZ0731: Ah I got it.

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: helperPoints get updated on test 7 and all of these tests run in consecutive order :)

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: Unfortunately the webdriver is very unpredictable and without the waits it errors out more often. The examples I saw online typically have like 30seconds of wait for things - which I thought was too much, but 5-10 seconds doesn't seem like that much. 

In this particular piece of code it really isn't waiting longer than necessary, it's just polling constantly checking to see if the alert has shown up yet (or is accessible by the driver), waiting a second and then trying again. If it succeeds, it breaks, but at the most it takes 60seconds before erroring out. Because the alerts are considered a different "window" of sorts by the browser, this type of while loop helps the webdriver effectively find the alert. At least that's what I found from some suggestions online and it seemed to help as I was running into more errors before doing this.

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: The wait here is to give time for the response to the click to 'process' as it sends the driver to another page.

src/main/webapp/user_profile.jsp
poncem91: I renamed the function on the homepage script file but not the user_profile script :)

src/main/webapp/homepage_script.js
poncem91: Ah yes, I forgot to edit the comment, I'll change this now, thanks!

src/main/webapp/homepage_script.js
poncem91: Sounds good!

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: I'm going to keep this one and the other one mentioned in the comment below as is, because of how they don't all behave the same, and making those exceptions in the for loop might make the code not readable at all.

src/test/java/com/google/neighborgood/IntegrationTest.java
poncem91: See above :)

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -310,36 +292,23 @@ public void _07_Userpage_AsLoggedUser_VerifyCompletedTask() {
...
+    String[] taskContentsTarget = {"overview", "helper", "status"};
+
+    // Iterates over task elements and compares the element's text with that of the stored
+    // recentTask
+    for (int i = 0; i < taskContentsTarget.length; i++) {
+      String taskContentItem =
+          wait.until(
+                  new Function<WebDriver, WebElement>() {
+                    public WebElement apply(WebDriver driver) {
+                      return driver.findElement(By.xpath(awaitVerifTaskXPath + "/td[1]"));
```
LeonardZ0731: Should this line be `awaitVerifTaskXPath + "/td[" + Integer.toString((i + 1)) + "]"))`?

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -256,46 +260,24 @@ public void _06_Userpage_AsLoggedHelper_CompleteTask() {
...
+
+    // Iterates over task elements and compares the element's text with that of the stored
+    // recentTask
+    for (int i = 0; i < taskContentsTarget.length; i++) {
+      String taskContentItem =
+          wait.until(
+                  new Function<WebDriver, WebElement>() {
+                    public WebElement apply(WebDriver driver) {
+                      return driver.findElement(
+                          By.xpath(taskCompletedXPath + "/td[" + (i + 1) + "]"));
```
LeonardZ0731: Personally I don't prefer adding an int directly with String, although it works : )

src/test/java/com/google/neighborgood/IntegrationTest.java
LeonardZ0731: Ah I saw you have fixed this in one commit.

src/test/java/com/google/neighborgood/IntegrationTest.java
LeonardZ0731: Also, since we add 1 to i during each loop, why not start the loop from i = 1?

poncem91: > Also, since we add 1 to i during each loop, why not start the loop from i = 1?

I thought about that too, but then in the assertEquals statement the array would have an index of [i-1] and the condition in the for loop would also have to have a 1 added, so either way I'd have to add or subtract 1. Personally I prefer this approach.

poncem91: > Personally I don't prefer adding an int directly with String, although it works : )

I prefer leaving it without explicitly converting it to a string since string concatenation in Java takes care of that for me :)
But I'll keep this is mind for when using other languages in the future!

src/main/webapp/user_profile.jsp
LeonardZ0731: Ah my fault. I thought it was the index.jsp 🤣

src/test/java/com/google/neighborgood/IntegrationTest.java
LeonardZ0731: Cool.

LeonardZ0731:
When I ran the test it also took really a long time and led to a time out at the end. I thought maybe I kept too many chrome tabs open when I ran the test. I will try again on Monday.

poncem91:
> When I ran the test it also took really a long time and led to a time out at the end. I thought maybe I kept too many chrome tabs open when I ran the test. I will try again on Monday.

I just remembered I also enabled hyperthreading on Chrome - which may have been a determining factor to speed things up. You can try that out here: chrome://flags/#scheduler-configuration

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1038 @@
...
+   * entities
+   */
+  public static void setupClass() {
+    WebDriverManager.chromedriver().setup();
+    driver = new ChromeDriver();
+    // FluentWait set to timeout after 60 seconds of waiting for a WebElement to be returned and
+    // polling every second
+    wait =
+        new FluentWait<WebDriver>(driver)
+            .withTimeout(Duration.ofSeconds(60))
```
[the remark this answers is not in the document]
poncem91: I can definitely shorten this up! I'll also look into ways of exiting the tests when it errors out/fails a test. I understand the frustration of waiting for a long time only for it to have failed from the beginning

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1038 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.junit.runners.MethodSorters;
+import org.openqa.selenium.*;
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.support.ui.FluentWait;
+import org.openqa.selenium.support.ui.Select;
+
+@RunWith(JUnit4.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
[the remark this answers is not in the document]
poncem91: Ah, yes I can add some comments and will look into making the tests less co-dependent. I thought setting up things in the background for some of the tests was almost like cheating since it wasn't actually testing out how the user would interact with it :) but this sounds like a simpler approach, so I'll start making some of these changes next week.

src/test/java/com/google/neighborgood/IntegrationTest.java
```diff
@@ -0,0 +1,1038 @@
...
+
+    js = (JavascriptExecutor) driver;
+    clearAllDatastoreEntities(driver);
+  }
+
+  @AfterClass
+  public static void teardown() {
+    if (driver != null) {
+      driver.quit();
+    }
```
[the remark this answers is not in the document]
poncem91: I don't believe so, none of the examples in their library repo had any additional `teardown`s and couldn't find anything that suggests otherwise

src/test/java/com/google/neighborgood/IntegrationTest.java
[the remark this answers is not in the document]
poncem91: I'll make the changes to remove the FixMethodOrder decorator in another PR as this one seems pretty long run already. Would it be alright to eliminate the clearAllDatastoreEntities method altogether and have testers run `mvn clean` before the test instead? There's been 2-3 instances where the test suite has failed on my end on the first test because all task entities weren't successfully deleted with webdriver, so perhaps eliminating this external factor from the test would be better.

Issue #84, opened 2020-07-15
Removal of the 400 redirects for all errors caused by empty input


LeonardZ0731:
let's open an issue to track removal of the 400 redirects for these errors

_Originally posted by @swparr in https://github.com/googleinterns/NeighborGood/pull/83_

PR #88, opened 2020-07-16
Address Issue #84 and remove all the 400 redirects caused by empty input

LeonardZ0731:
Update three servlet files to remove all the 400 redirects.
Update account.jsp, index.jsp and user_profile.jsp to add onsubmit function to all the forms. The onsubmit function will check whether the input is empty or empty spaces before submitting the form. If they are empty spaces, show an alert box and mark the required areas as red.
Modify css files to accomodate the change and add the onsubmit function to the script file


LeonardZ0731:
The latest commit accomodates the change made in Mafe's PR #90, indicating that placing a marker on the map to set the address is mandatory.

src/main/webapp/user_profile_script.js
```diff
@@ -18,6 +18,48 @@ const GOOGLE_KIRKLAND_LAT = 47.669846;
...
+        }
+    }
+    if (!result) {
+        alert("All fields are required. Please fill out all fields with non-empty input.");
+        return false;
+    }
+    return true;
+}
+
+function validateInfoForm(id) {
```
poncem91, line 42: Ah this actually works better than the lat/lng validation I did on my PR for the GeoQueries. I'll wait until your PR is merged and I can use this function also to validate that those two fields aren't empty. Looks good though!

PR #89, opened 2020-07-16
remove userId property from UserInfo entity to avoid redudancy

### LINKED ISSUE #67: remove userId property from UserInfo entities as the entity id now holds the same value

poncem91:
fixes #67

### DIFF
--- a/src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
+++ b/src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
@@ -17,10 +17,9 @@ package com.google.neighborgood.helper;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
-import com.google.appengine.api.datastore.PreparedQuery;
-import com.google.appengine.api.datastore.Query;
-import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.users.UserService;
 import java.util.ArrayList;
 import java.util.List;
@@ -28,13 +27,11 @@ import java.util.List;
 public final class RetrieveUserInfo {
   public static List<String> getInfoFromId(String userId) {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Query query =
-        new Query("UserInfo")
-            .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, userId));
-    PreparedQuery results = datastore.prepare(query);
-
-    Entity entity = results.asSingleEntity();
-    if (entity == null) {
+    Key userEntityKey = KeyFactory.createKey("UserInfo", userId);
+    Entity entity;
+    try {
+      entity = datastore.get(userEntityKey);
+    } catch (EntityNotFoundException e) {
       return null;
     }
 
--- a/src/test/java/com/google/neighborgood/UserInfoServletTest.java
+++ b/src/test/java/com/google/neighborgood/UserInfoServletTest.java
@@ -185,9 +185,10 @@ public final class UserInfoServletTest {
     try {
       entity = ds.get(userEntityKey);
     } catch (EntityNotFoundException e) {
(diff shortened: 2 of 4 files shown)


LeonardZ0731:
Have you tested on the functionality after making the change? I'm a little bit worrid about creating a key using id instead of keyString.

LeonardZ0731:
Ah I think that works.

poncem91:
> Have you tested on the functionality after making the change? I'm a little bit worrid about creating a key using id instead of keyString.

Yup! Ran all the tests and tested it out myself, but let me know if you find any bugs because of it.

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -181,14 +180,14 @@ public void normalMultipleInputTest() throws IOException {
...
-
-    // The entity can't be null
-    assertNotNull(entity);
+    Key userEntityKey = KeyFactory.createKey("UserInfo", "1234567890");
+    Entity entity = null;
+    try {
+      entity = ds.get(userEntityKey);
+    } catch (EntityNotFoundException e) {
+      // The entity can't be null
+      assertNotNull(entity);
```
LeonardZ0731, line 189: I think the assertNotNull statement is in the wrong place?

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -206,14 +205,13 @@ public void normalMultipleInputTest() throws IOException {
...
-                .setFilter(new FilterPredicate("userId", FilterOperator.EQUAL, "1234567890")));
-    entity = results.asSingleEntity();
 
-    // The entity can't be null
-    assertNotNull(entity);
+    try {
+      entity = ds.get(userEntityKey);
+    } catch (EntityNotFoundException e) {
+      // The entity can't be null
+      assertNotNull(entity);
```
LeonardZ0731, line 213: Same problem here. Maybe the assertNotNull statement should be placed out of the catch clause

src/test/java/com/google/neighborgood/UserInfoServletTest.java
```diff
@@ -309,14 +307,16 @@ public void emptyInputTest() throws IOException {
     // The entity can't be null
     assertNotNull(entity);
 
+    System.out.println("\n\n\nin test. user id is: " + entity.getKey().getName());
```
LeonardZ0731, line 310: I think this print statement could be deleted if it is no longer needed.

poncem91: Oh yes, I forgot to remove this, thanks for pointing this out!

src/test/java/com/google/neighborgood/UserInfoServletTest.java
poncem91: I'll move it outside of the catch statement :)

src/test/java/com/google/neighborgood/UserInfoServletTest.java
poncem91: Got it, I can move it to outside of the catch statement.

src/main/java/com/google/neighborgood/helper/RetrieveUserInfo.java
```diff
@@ -17,24 +17,21 @@
...
-import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.users.UserService;
 import java.util.ArrayList;
 import java.util.List;
 
 public final class RetrieveUserInfo {
```
peap, line 27: Can you add some unit tests for this class?

poncem91: Yep! Should I go ahead and add these into this PR or create a new one for it?

LeonardZ0731: Ahh I have already added a test case for this class in PR #79

LeonardZ0731: In PR #79 there is a java file called `HelperFunctionTest.java` that tested the `RetrieveUserInfo` class and `RewardingPoints` class

poncem91: Ah, got it!

PR #91, opened 2020-07-18
change task list look and fix some glitchy behavior with screen resizing

### LINKED ISSUE #86: make homepage task click functionality changes for detail modal
I would prefer to have this onclick function dynamically added through the script file. Another suggestion would be to have the whole task div be clickable (sort of like the user tasks in the admin page). I'd like to work on this latter modification since I was planning on doing same changes to the homepage task list anyway, so I can go ahead and change the onclick part as well if you like.

_Originally posted by @poncem91 in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjg0MzM2NTEwOnYy/pull_request_review_threads/discussion_

poncem91:
This also resolves #86

### DIFF
--- a/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/TaskServlet.java
@@ -29,6 +29,8 @@ import com.google.gson.Gson;
 import com.google.neighborgood.helper.RetrieveUserInfo;
 import com.google.neighborgood.helper.RewardingPoints;
 import java.io.IOException;
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


poncem91:
> LGTM! Adding timestamp for the task is really a good idea and replacing onclick with addEventListener in the script file is also great! But I have a question: Can you explain why we need to have all these `e.stopPropagation();`? I am a little bit confused about the use of these lines of code.

😄  - e.stopPropagation() basically stops the click event from bubbling up the parent divs. So when I click for example the Help Out button, it will trigger the helpOut() funtion but it won't trigger its parent on click function showTaskInfo() (it's considered a parent click event because the help-button is inside the task div that holds the showTaskInfo() click event).

PR #92, opened 2020-07-20
add option to let users choose what distance radius to query

poncem91:


src/main/java/com/google/neighborgood/servlets/TaskServlet.java
```diff
@@ -65,18 +66,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Location coordinates are missing");
     }
 
+    if (request.getParameterMap().containsKey("miles")) {
+      double miles = 5;
```
LeonardZ0731, line 70: Can we define a static final variable `DEFAULT_MILE_RADIUS` and set it to 5? This will make the code more understandable

poncem91: Good idea, I'll make these changes, thanks!

swparr:
Should we have the drop down on the page when there are no tasks in the system? I only see it displayed once a tasks exists. (This can be done under a separate PR/Issue)

poncem91:
> Should we have the drop down on the page when there are no tasks in the system? I only see it displayed once a tasks exists. (This can be done under a separate PR/Issue)

This is a good point. I will add an issue for this to modify in a next PR, thanks!

PR #93, opened 2020-07-20
Add a chat box to the task detail window of the user profile page

LeonardZ0731:
Create a MessageServlet.java file to handle the GET and POST request sent to /messages. If it receives a POST request, it will store the new message in the datastore along with the sender information. If it receives a GET request, it will return a list of messages related with the task.
Modify the user_profile.jsp to add the chat box and a div element to hold all the messages. Update the css file to accomodate the change.
Update the script file to add a new function loadMessages(keyString) that will add all messages returned by the response of GET request to the message container

### DIFF
--- a/src/main/java/com/google/neighborgood/servlets/MessageServlet.java
+++ b/src/main/java/com/google/neighborgood/servlets/MessageServlet.java
@@ -1,0 +1,109 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.neighborgood.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.gson.Gson;
+import com.google.neighborgood.Message;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that loads and records new message entities. */
+@WebServlet("/messages")
+public class MessageServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String taskKey = request.getParameter("key");
+    if (taskKey == null) {
+      System.err.println("No task key string provided");
+    }
+
+    // Make sure that the user has already logged into his account
+    UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect(userService.createLoginURL("/account.jsp"));
+      return;
+    }
+
+    Filter filter = new FilterPredicate("taskId", FilterOperator.EQUAL, taskKey);
+    Query query =
+        new Query("Message").setFilter(filter).addSort("sentTime", SortDirection.ASCENDING);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    List<Message> messages = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+      messages.add(new Message(entity));
+    }
+
+    Gson gson = new Gson();
+    String json = gson.toJson(messages);
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // First check whether the user is logged in
+    UserService userService = UserServiceFactory.getUserService();
+
+    if (!userService.isUserLoggedIn()) {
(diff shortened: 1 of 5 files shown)


src/main/webapp/user_profile_style.css
```diff
@@ -294,3 +294,29 @@ table button:hover i {
...
+  border: 2px solid #dedede;
+  background-color: #f1f1f1;
+  border-radius: 5px;
+  padding: 10px;
+  margin: 10px 10px;
+}
+
+.sentByMe {
+  color: white;
+  background-color: #1aff1a;
```
poncem91, line 316: nit: could we use a less bright green for this? :)

src/main/webapp/user_profile.jsp
```diff
@@ -220,6 +220,21 @@
             <span class="close-button" id="task-info-close-button" onclick="closeTaskInfoModal()">&times;</span>
             <h1>Task Detail: </h1>
             <div id="task-detail-container"></div>
+            <hr/>
+            <h1>Chat: </h1>
+            <div id="message-container"></div>
+            <form id="chat-box" action="/messages" method="POST" onsubmit="return validateMessage()">
```
poncem91, line 226: This is throwing an error `Uncaught ReferenceError: validateMessage is not defined` - I think the function might be missing from the script file

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // Make sure that the user has already logged into his account
+    UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect(userService.createLoginURL("/account.jsp"));
+      return;
+    }
+
+    Filter filter = new FilterPredicate("taskId", FilterOperator.EQUAL, taskKey);
```
poncem91, line 55: nit: this wording is a bit confusing, is it the id or the key?

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    Query query =
+        new Query("Message").setFilter(filter).addSort("sentTime", SortDirection.ASCENDING);
```
poncem91, line 57: Would it be a good idea to store the messages already queried (perhaps on a http session) and only query new ones so that we're not querying the same messages from the database everytime we open the task details modal?

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -0,0 +1,109 @@
...
+      System.err.println("The input message is empty");
+      return;
+    }
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Create a message entity to store the related information
+    Entity msgEntity = new Entity("Message");
+    msgEntity.setProperty("message", message);
+    msgEntity.setProperty("taskId", keyString);
```
poncem91, line 102: Same comment as before. I would suggest wording this as either id or key, instead of using both word choices.

src/main/webapp/user_profile_script.js
```diff
@@ -60,6 +60,24 @@ function validateInfoForm(id) {
     return true;
 }
 
+async function loadMessages(keyString) {
+    console.log("Start loading");
```
poncem91, line 64: You can probably remove this console.log line, unless it was left here on purpose :)

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    String message = request.getParameter("msg");
+    if (message == null || message.trim().equals("")) {
+      System.err.println("The input message is empty");
+      return;
+    }
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Create a message entity to store the related information
+    Entity msgEntity = new Entity("Message");
```
poncem91, line 100: I would suggest you look into parent entities. We used it for declaring Users as parents of Tasks. Here we could do something similar, declaring Tasks as parents of Messages. If a Message had a Task as a parent, you could do `messageEntity.getParent()` and it would give you the parent task's key. 
There's also ancestor queries, which I think would be useful in the doGet portion of this sevlet.
But it's totally up to you, this also works!

https://cloud.google.com/appengine/docs/standard/java/datastore/queries#ancestor_queries
https://cloud.google.com/appengine/docs/standard/java/datastore/entities#Ancestor_paths

src/main/webapp/user_profile_style.css
LeonardZ0731: Sure. I will change to a darker green in the new commit.

src/main/webapp/user_profile.jsp
LeonardZ0731: Ah I forgot to add it when I commit the new changes. I will include the function in the new commit.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: Ah that would be a really challenging part, since we need to create a session for every single task. Managing these sessions would be really complicated, and storing all these messages could be space inefficient.

poncem91: It could just be one session per user, but it's all good as is for this :)

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: I would use id for these variables in the next commit.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: I apologized for the confusion brought by the wording. I will use id for all the related variables.

src/main/webapp/user_profile_script.js
LeonardZ0731: Ah I will remove this in the new commit.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: I would consider adding this parent entity feature in a new PR and I've created issue #94 for this.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    Gson gson = new Gson();
+    String json = gson.toJson(messages);
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
dandrews-code, line 74: Just to confirm, can anyone send a message for any task?  Do we need to confirm that this task is in the user's neighborhood, or that they are owner/assigned the task?

LeonardZ0731: Currently the form to send an message is only available on the user profile page. A task will only show on the user profile page if the user is the owner or helper of the task.

PR #107, opened 2020-07-24
change implementation of pagination to infinite-scroll by using cursors and function on a load-more basis when scrolling to the bottom

### LINKED ISSUE #104: make use of cursors in the pagination
now that we will now be using geospatial queries, we can use cursors in the pagination implementation which will better performance and will result in a better solution when scaling

poncem91:
Resolves #104


poncem91:
> Will you add similar scrolling to the 'my tasks' page. We should open an issue for that

I'll talk to Leonard about this so we can open an issue and tackle this if time permits.

PR #109, opened 2020-07-24
Address the scaling issue of the message entities

LeonardZ0731:
- Add a feature that at most 10 messages get loaded each time. If the user wants to check previous messages, they can scroll to the top and click the button to load 10 more messages. In this way, only a reasonable amount of messages get loaded on the screen each time.
- Decrease the number of doGet calls to improve the efficiency of sending messages. Now, the doGet function is not called every time the user send a new message. Instead, a div element will be created to hold the message and add it to the end of the chat box.
- Add a feature that messages related with a task will be deleted once the task has been completed, abandoned or deleted to remove all unnecessary message entities from datastore in a timely manner.


src/main/webapp/user_profile_script.js
```diff
@@ -83,6 +89,8 @@ async function refresh() {
     loadMessages(document.getElementById("chat-id-input").value);
 }
 
+// To address the scaling issue and avoid loading too many messages at a time,
+// the chat box would first load at most 10 most recent messages.
```
poncem91, line 93: It seems like you are still fetching all the messages related to the task but only displayed the 10 most recent ones. To address the scaling issue, would it be better to only retrieve 10 messages from datastore at a time?

src/main/webapp/user_profile_script.js
```diff
@@ -91,18 +99,88 @@ async function loadMessages(keyString) {
...
+// task with keyString equals to key. After loading, scroll to element messageDiv
+async function loadMoreMessages(key, messageDiv) {
+    const queryURL = "/messages?key=" + key;
+    const request = new Request(queryURL, {method: "GET"});
+    const response = await fetch(request);
+    const msgResponse = await response.json();
+
+    const msgContainer = document.getElementById("message-container");
+    msgContainer.removeChild(msgContainer.childNodes[0]);
+    var count = msgContainer.childNodes.length + 10;
```
poncem91, line 146: Same comment as above. It might be worth looking into using Cursors in Datastore to reduce the number of message entities retrieved while keeping track of where you are on the query :)

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -106,4 +107,38 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // Get the task ID
+    String taskId = request.getParameter("task-id");
+    if (taskId == null) {
+      System.err.println("The task id is not included");
+      return;
+    }
+
+    // Get all stored entities related with the task that corresponds to the given taskId
+    Filter filter = new FilterPredicate("taskId", FilterOperator.EQUAL, taskId);
+    Query query = new Query("Message").setFilter(filter);
```
poncem91, line 131: I don't know how much of a difference this could make, but Datastore has a special type of query to only returns entity's keys which is supposed to be lower-latency, so if you're thinking about scaling issues it might be worth looking into this as well: https://cloud.google.com/appengine/docs/standard/java/datastore/queries#keys-only_queries

poncem91:
I'm getting this error on the terminal window when loading the userpage: `Unable to find the owner of the task in the database` 
I think it stems from the Task.java file as I ran into a similar message on a recent PR and had to make some changes for it. I believe it's been fixed already in the master branch but just a heads up!

src/main/webapp/user_profile_script.js
LeonardZ0731: Ah, that's a really good idea. I will use Cursor to solve this problem in the next commit.

src/main/webapp/user_profile_script.js
LeonardZ0731: Sure. I would use a Cursor entity to keep track of the last retrieved message entity.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: Cool! I will try to use that in the new commit.

LeonardZ0731:
> I'm getting this error on the terminal window when loading the userpage: `Unable to find the owner of the task in the database`
> I think it stems from the Task.java file as I ran into a similar message on a recent PR and had to make some changes for it. I believe it's been fixed already in the master branch but just a heads up!

I didn't get this error when testing on my branch. Maybe it's related with some tasks you created in datastore before merging the PR that makes the User entity a parent entity of task entity?

src/main/webapp/user_profile_script.js
```diff
@@ -83,26 +90,101 @@ async function refresh() {
...
+    const result = await response.json();
+    const msgResponse = result.messages;
+    cursorString = result.cursorString;
 
     const msgContainer = document.getElementById("message-container");
     msgContainer.innerHTML = "";
-    for (var index = 0; index < msgResponse.length; index++) {
-        var msg = msgResponse[index];
+    if (msgResponse.length === 10) {
+        var msg = msgResponse[9];
```
claudiatu, line 106: Similar to my other review, is `var` necessary here?

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
```diff
@@ -106,4 +133,38 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    }
+
+    // Get the task ID
+    String taskId = request.getParameter("task-id");
+    if (taskId == null) {
+      System.err.println("The task id is not included");
+      return;
+    }
+
+    // Get all stored entity ke related with the task that corresponds to the given taskId
```
claudiatu, line 155: nit: key

src/main/webapp/user_profile_script.js
LeonardZ0731: I don't think they are necessary here🤣 I would change it to `let` in the new commit.

src/main/java/com/google/neighborgood/servlets/MessageServlet.java
LeonardZ0731: Sure. I would fix it in the new commit.

src/main/webapp/user_profile_script.js
```diff
@@ -83,26 +90,101 @@ async function refresh() {
...
+        for (var index = 0; index < 10; index++) {
+            var message = msgResponse[index];
+            newMessageDiv = document.createElement("div");
+            newMessageDiv.className = message.className;
+            newMessageDiv.appendChild(document.createTextNode(message.message)); 
+            msgContainer.insertBefore(newMessageDiv, msgContainer.childNodes[0]);
+        }
+        var moreMessageBtn = document.createElement("button");
+        moreMessageBtn.setAttribute("class", "more-msg-button");
+        moreMessageBtn.textContent = "Load 10 more messages";
```
swparr, line 162: I'd reword this to 'Load 10 previous messages' or Load older messages

LeonardZ0731: Sure. "Load 10 previous messages" is better than my present wording and I would make the change in a new PR.

PR #111, opened 2020-07-27
add searchbox to load tasks by different neighborhood

poncem91:
An autocomplete search box was added to the top of the task results that allows users to enter a different address, select it and this will load tasks by the addresses's neighborhood instead. Additionally, if the search box doesn't have a valid input (a selection hasn't been made from the drop-down menu) it will default to the user's actual location.

### DIFF
--- a/src/main/webapp/homepage_script.js
+++ b/src/main/webapp/homepage_script.js
@@ -439,18 +439,18 @@ function fetchTasks(category, cursor) {
 function displayTasks(append) {
     if (taskGroup !== null && taskGroup.currentTaskCount > 0) {
         document.getElementById("no-tasks-message").style.display = "none";
-        document.getElementById("loading").style.display = "none";
         document.getElementById("tasks-message").style.display = "block";
         if (append) document.getElementById("tasks-list").innerHTML += taskGroup.tasks;
         else document.getElementById("tasks-list").innerHTML = taskGroup.tasks;
         document.getElementById("tasks-list").style.display = "block";
         addTasksClickHandlers();
     } else {
-        document.getElementById("loading").style.display = "none";
         document.getElementById("no-tasks-message").style.display = "block";
         document.getElementById("tasks-message").style.display = "none";
         document.getElementById("tasks-list").style.display = "none";
     }
+    document.getElementById("loading").style.display = "none";
+    document.getElementById("search-box").style.visibility = "visible";
 }
 
 /* Function adds all the necessary tasks 'click' event listeners*/
--- a/src/main/webapp/homepage_script.js
+++ b/src/main/webapp/homepage_script.js
@@ -318,15 +318,30 @@ begins the processes of retrieving the user's location*/
 function getTasksForUserLocation() {
     const script = document.createElement("script");
     script.type = "text/javascript";
-    script.src =  "https://maps.googleapis.com/maps/api/js?key=" + MAPSKEY + "&callback=initialize&language=en";
+    script.src =  "https://maps.googleapis.com/maps/api/js?key=" + MAPSKEY + "&callback=initialize&libraries=places&language=en";
     script.defer = true;
     script.async = true;
     document.head.appendChild(script);
-	
+
     // Once the Maps API script has dynamically loaded it gets the user location,
     // waits until it gets an answer updates the global userLoaction variable and then calls
     // fetchTasks and displayTasks
 	window.initialize = function () {
+        let placeAutocomplete = new google.maps.places.Autocomplete(document.getElementById("place-input"));
+        placeAutocomplete.setFields(['geometry']);
+        google.maps.event.addListener(placeAutocomplete, 'place_changed', function() {
+                let place = placeAutocomplete.getPlace();
+                userLocation = place.geometry.location.toJSON();
+                toNeighborhood(userLocation)
+                    .then(() => fetchTasks())
+                    .then(response => {
+                            displayTasks(response);
+                        })
+                    .catch(() => {
+                        console.error("User location and/or neighborhood could not be retrieved");
+                        document.getElementById("location-missing-message").style.display = "block";
+                    });
+              });
          getUserLocation().then(location => toNeighborhood(location))
         	.then(() => fetchTasks(currentCategory))
             .then(response => {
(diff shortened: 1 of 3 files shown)


src/main/webapp/homepage_script.js
```diff
@@ -307,15 +308,35 @@ begins the processes of retrieving the user's location*/
...
+                    userLocation = userActualLocation;
+                }
+                toNeighborhood(userLocation)
+                        .then(() => fetchTasks(currentCategory, "clear"))
+                        .then(response => {
+                                taskGroup = response;
+                                displayTasks(response);
+                            })
+                        .catch(() => {
+                            console.error("User location and/or neighborhood could not be retrieved");
```
LeonardZ0731, line 336: If the user location or neighborhood could not be retrieved, do we have to report an error in the console? I think maybe just provide the users with a proper method to select their own positions could be a better solution.

src/main/webapp/homepage_script.js
```diff
@@ -410,18 +434,18 @@ function fetchTasks(category, cursorAction) {
 function displayTasks(append) {
     if (taskGroup !== null && taskGroup.currentTaskCount > 0) {
         document.getElementById("no-tasks-message").style.display = "none";
-        document.getElementById("loading").style.display = "none";
         document.getElementById("tasks-message").style.display = "block";
-        if (append) document.getElementById("tasks-list").innerHTML += taskGroup.tasks;
+        if (append == true) document.getElementById("tasks-list").innerHTML += taskGroup.tasks;
```
LeonardZ0731, line 438: Is there a reason for using `append == true` here? I think `if (append)` should perform the same and is better in style.

src/main/webapp/homepage_script.js
poncem91: Ah, you're right! I'll remove the console error message and update the message displayed to the user in the site to reflect the search box changes, thanks!

src/main/webapp/homepage_script.js
poncem91: I previously had a different implementation where this wasn't working, but then I changed the implementation and forgot to change this statement. I'll change this in the next commit, thanks!

src/main/webapp/index.jsp
```diff
@@ -105,13 +105,17 @@
...
+                <input id="place-input" name="place-input" type="text" placeholder=" Type an address here" class="pac-target-input" autocomplete="off">
+              </div>
 
               <!--Results Messages-->
               <div id="loading" class="results-message">
                   <img id="loading-gif" src="images/loading.gif" alt="Loading..."/>
               </div>
               <div id="location-missing-message" class="results-message">
-                  We could not retrieve your location to display your neighborhood tasks.
+                  We could not retrieve your location to display your neighborhood tasks. Please use the search box to manually enter a location.
```
claudiatu, line 118: 🎉

src/main/webapp/index.jsp
```diff
@@ -105,13 +105,17 @@
                   }
                   %>
               </div>
+              <div id="search-box">
+                <label for="place-input">Search tasks in a different location:</label>
+                <input id="place-input" name="place-input" type="text" placeholder=" Type an address here" class="pac-target-input" autocomplete="off">
```
claudiatu, line 110: What does `pac` in `pac-target-input` mean in this context?

src/main/webapp/homepage_script.js
```diff
@@ -307,23 +308,42 @@ begins the processes of retrieving the user's location*/
...
     script.type = "text/javascript";
-    script.src =  "https://maps.googleapis.com/maps/api/js?key=" + MAPSKEY + "&callback=initialize&language=en";
+    script.src =  "https://maps.googleapis.com/maps/api/js?key=" + MAPSKEY + "&callback=initialize&libraries=places&language=en";
     script.defer = true;
     script.async = true;
     document.head.appendChild(script);
-	
+
     // Once the Maps API script has dynamically loaded it gets the user location,
     // waits until it gets an answer updates the global userLoaction variable and then calls
```
claudiatu, line 317: nit: userLocation

claudiatu, line 317: Should this comment be moved below, or updated?

src/main/webapp/homepage_script.js
```diff
@@ -307,23 +308,42 @@ begins the processes of retrieving the user's location*/
...
     // fetchTasks and displayTasks
 	window.initialize = function () {
+        let placeAutocomplete = new google.maps.places.Autocomplete(document.getElementById("place-input"));
+        placeAutocomplete.setFields(['geometry']);
```
claudiatu, line 321: suggestion: add a comment saying that the geometry field indicates the location (lat/long) and viewport for the returned data

src/main/webapp/homepage_script.js
```diff
@@ -410,18 +433,18 @@ function fetchTasks(category, cursorAction) {
...
         else document.getElementById("tasks-list").innerHTML = taskGroup.tasks;
         document.getElementById("tasks-list").style.display = "block";
         addTasksClickHandlers();
     } else {
-        document.getElementById("loading").style.display = "none";
         document.getElementById("no-tasks-message").style.display = "block";
         document.getElementById("tasks-message").style.display = "none";
         document.getElementById("tasks-list").style.display = "none";
     }
+    document.getElementById("loading").style.display = "none";
```
claudiatu, line 446: nice 😀

src/main/webapp/index.jsp
poncem91: Place Autocomplete, however I just realized I forgot to remove this class attribute here as I am no longer using it :) thanks for pointing this out!

src/main/webapp/homepage_script.js
poncem91: Woops, will update this, thanks!

poncem91: Yes, I'll update this comment to include the most recent changes to the initialize function. Thanks!

src/main/webapp/homepage_script.js
poncem91: Good idea! I'll include this in the next commit, thanks!

src/main/webapp/homepage_script.js
poncem91: Thanks 😄 !

poncem91:
> Sorry for the late reply!

No worries at all, thanks for the feedback :) !

PR #112, opened 2020-07-27
Test cases for the newly-added MessageServlet.java

LeonardZ0731:


src/test/java/com/google/neighborgood/MessageServletTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testEnvironmentTest() {
+    // Test the UserService feature
+    assertTrue(userService.isUserAdmin());
```
swparr, line 84: Why do we assert the user is an admin? Are there any differences in how this works between an admin and a regular user?

LeonardZ0731: Actually there are no differences between an admin and a regular user. The assertion here actually just makes sure that the test environment has been set up correctly. When I set up the test environment, I made the mock user as the admin, so I wanted to make sure that the setup worked properly here.

PR #117, opened 2020-07-29
implement map view option

### LINKED ISSUE #29: POST MVP: Option to toggle between Map-view and List-view for homepage

### LINKED ISSUE #77: Centralize the map creation process
Does this map use the same styles as the other maps in the site? Can we centralize the map creation somewhere after this PR?

_Originally posted by @peap in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjgzMTcxNDA2OnYy/pull_request_review_threads/discussion_

poncem91:
- Changes implementation of TaskServlet's doGet response so response can both be used for list view as well as map view (consequently a new createTaskListNode function was needed in the script file).
- Adds buttons to UI to toggle between map and list view.
- Loads markers per every task loaded into map (markers with exact same lat/lng coordinates will be on top of each other so open click these open up like a spider using a library I found).
- Load more button shows up if there are more tasks in the query that haven't been returned - upon pressing this, 10 more tasks are loaded into the map.
- Upon drag into a different location, the map checks if the new map center is in a different neighborhood, if so it shows the search this neighborhood control button, and upon clicking it it fetches tasks from that neighborhood.
- Markers can be clicked to see the same information available in list-view and the infowindow that opens up can be clicked to open up the task detail modal.

Resolves #29 
Resolves #77 by making use of an external map styling file.

### DIFF
--- a/src/main/java/com/google/neighborgood/helper/TaskGroup.java
+++ b/src/main/java/com/google/neighborgood/helper/TaskGroup.java
@@ -21,6 +21,7 @@ import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
+import com.google.neighborgood.User;
 import com.google.neighborgood.task.Task;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -30,8 +31,8 @@ import java.util.HashMap;
  * or not the end of the query has been reached
  */
 public class TaskGroup {
-  private static HashMap<String, String>
-      usersNicknames; // Stores task owner's user info to prevent querying multiple times in
+  private static HashMap<String, User>
+      usersInfo; // Stores task owner's user info to prevent querying multiple times in
   // datastore for the same user's info
   private static DatastoreService datastore;
   private final boolean userLoggedIn;
@@ -40,7 +41,7 @@ public class TaskGroup {
   private ArrayList<Task> tasks;
 
   public TaskGroup() {
-    this.usersNicknames = new HashMap<String, String>();
+    this.usersInfo = new HashMap<String, User>();
     UserService userService = UserServiceFactory.getUserService();
     this.userLoggedIn = userService.isUserLoggedIn();
     this.datastore = DatastoreServiceFactory.getDatastoreService();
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/neighborgood/helper/TaskGroup.java
```diff
@@ -19,101 +19,65 @@
...
 import java.util.HashMap;
 
 /**
  * Helper class that stores the HTML depiction of tasks in groups of 10 or less along with whether
  * or not the end of the query has been reached
  */
 public class TaskGroup {
-  private static HashMap<String, String>
-      usersNicknames; // Stores task owner's user info to prevent querying multiple times in
+  private static HashMap<String, User>
```
LeonardZ0731, line 34: nit: I would suggest declaring the type as Map instead of HashMap since we can change the underlying object to be a different kind of map without breaking any part of the code. If we declare it as HashMap and wrote some code that depends on the feature that is specific to HashMap, we may have to change the contract for TaskGroup class after changing the implementation.

src/main/java/com/google/neighborgood/helper/TaskGroup.java
```diff
@@ -19,101 +19,65 @@
...
+      usersInfo; // Stores task owner's user info to prevent querying multiple times in
   // datastore for the same user's info
   private static DatastoreService datastore;
   private final boolean userLoggedIn;
-  private final String userId;
   private int currentTaskCount;
-  private StringBuilder tasks; // String-like representation of 10 tasks
   private boolean endOfQuery;
+  private ArrayList<Task> tasks;
```
LeonardZ0731, line 41: Same suggestion here. I would prefer `List<Task> tasks` to `ArrayList<Task> tasks`

src/main/java/com/google/neighborgood/helper/TaskGroup.java
```diff
@@ -19,101 +19,65 @@
...
-    }
-    tasks.append("</div>");
-    if (this.userLoggedIn) {
-      // changes the Help Button div if the current user is the owner of the task
-      if (!this.userId.equals(taskOwner)) {
-        tasks.append("<div class='help-out'>HELP OUT</div>");
-      } else {
-        tasks.append(
-            "<div class='help-out disable-help' title='This is your own task'>HELP OUT</div>");
+        taskOwnerNickname = "Neighbor";
```
LeonardZ0731, line 76: nit: How about changing the nickname to "Your Friendly Neighbor"? 😄

src/main/java/com/google/neighborgood/helper/TaskGroup.java
poncem91: Sounds good, I will make these changes! :)

src/main/java/com/google/neighborgood/helper/TaskGroup.java
poncem91: Will update this in the next commit, thanks!

src/main/java/com/google/neighborgood/helper/TaskGroup.java
poncem91: I like that idea! 😃

PR #126, opened 2020-07-31
Group web file

denzilbilson:


src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,41 @@
...
+        <web-resource-collection>
+            <web-resource-name>Admin</web-resource-name>
+            <url-pattern>/admin_dashboard.html</url-pattern>
+        </web-resource-collection>
+        <auth-constraint>
+            <role-name>admin</role-name>
+        </auth-constraint>
+    </security-constraint>
+
+    <security-constraint>
```
poncem91, line 27: Might be worth noting for reviewers that the HTTPS and error code portions only work on the deployed site.

denzilbilson: Yes, that is absolutely correct.

denzilbilson:
In case no one sees Mafe's comment above, many of the above actions in the web.xml do not work unless you are on a live, deployed server.

PR #130, opened 2020-08-04
Denzil admin prototype

denzilbilson:


src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,231 +1,264 @@
...
+
+async function openWithPopup(id){
+	const task = await searchTasks(id);
+	document.getElementById("task-detail-input").value = task.detail;
+	document.getElementById("reward-input").value = task.reward;
+	document.getElementById("task-id-input").value = task.keyString;
+	document.getElementById("edit-category-input").value = task.category;
+}
+
+async function openWithAdmin(id){
```
poncem91, line 250: Would it be better to add a delete icon like on the user tasks so that we know that clicking on that means delete and not edit, like on the user tasks?

src/main/webapp/admin_dashboard_script.js
```diff
@@ -1,231 +1,264 @@
...
+			adminTasksArray = tasks;
+			let taskSection = document.getElementById("admin-tasks");
+			for (adminTask of tasks) {
+				taskSection.innerHTML += addAdminTask(adminTask);
+            }
+		});
+}
+
+function addTask(task) {
+	let string = `<a href="#user-popup"><li class="admin-user-task"  onclick="openWithPopup('${task.keyString}')"><span><h3> ${task.category} </h3>`;
```
poncem91, line 214: Personally, I think that if the task changes when the mouse is hovered, it gives the feel that the whole task is a button, so it would make sense that if I click wherever on the task (except the delete button) the edit task modal would pop up.

src/main/java/com/google/neighborgood/task/adminTask.java
```diff
@@ -0,0 +1,28 @@
+package com.google.neighborgood.task;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.KeyFactory;
+
+public final class adminTask {
+  private final String keyString;
+  private final String detail;
+  private final String date;
+  private final String time;
```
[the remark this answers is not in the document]
denzilbilson: The html side makes sure that it is rendered in the time format so a string works best in this scenario

src/main/webapp/admin_dashboard.html
```diff
@@ -1,127 +1,93 @@
...
+					</div>
+				</div>
+			</main>
+		</div>
+		<div class="popup-overlay" id="user-popup">
+			<div id="popup">
+				<h2>Edit Task</h2>
+				<form action="/admin-user-tasks" method="POST">
+					Task Details: <input type="text" id="task-detail-input" name="task-detail-input">
+					Reward: <input type="text" id="reward-input" name="reward-input" min="0" max="200">
```
LeonardZ0731, line 64: I would suggest adding a onsubmit function to check whether the input task info field is empty or not. We add this new feature to the form submission process on the user profile page recently and it could prevent the users from submitting empty task info or task overview.

src/main/webapp/admin_dashboard.html
```diff
@@ -1,127 +1,93 @@
...
+		<link rel="stylesheet" href="admin_dashboard_style.css">
+		<script src='https://kit.fontawesome.com/a076d05399.js'></script>
+		<script src="https://www.gstatic.com/charts/loader.js"></script>
+		<script type="text/javascript" src="config.js"></script>
+		<script src="admin_dashboard_script.js"></script>
+	</head>
+	<body>
+		<div class="wrapper">
+			<main>
+				<h1><a href="/"><i class="fas fa-user-circle fa-2x" title="Go to User Page" aria-hidden="true"></i></a> Admin Dashboard</h1>
```
LeonardZ0731, line 30: Currently the user button will take the users back to the main page instead of user profile page, which is a little bit frustrating. Although I think it is a good idea to take the users back to the main page, we should have a different icon and add some padding to the top. Currently the icon is kinda too close to the top.

PR #132, opened 2020-08-05
change user's nickname's cache from a local hashmap to memcache

poncem91:
By using memcache the user's nicknames are stored in a distributed in-memory data cache instead of a temporary hashmap.

### DIFF
--- a/src/main/java/com/google/neighborgood/helper/TaskGroup.java
+++ b/src/main/java/com/google/neighborgood/helper/TaskGroup.java
@@ -19,23 +19,22 @@ import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.memcache.ErrorHandlers;
+import com.google.appengine.api.memcache.MemcacheService;
+import com.google.appengine.api.memcache.MemcacheServiceFactory;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.neighborgood.data.Task;
-import com.google.neighborgood.data.User;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/neighborgood/helper/TaskGroup.java
```diff
@@ -19,31 +19,31 @@
...
   private static DatastoreService datastore;
   private final boolean userLoggedIn;
   private int currentTaskCount;
   private boolean endOfQuery;
   private List<Task> tasks;
 
   public TaskGroup() {
-    this.usersInfo = new HashMap<String, User>();
+    this.syncCache = MemcacheServiceFactory.getMemcacheService();
+    syncCache.setErrorHandler(ErrorHandlers.getConsistentLogAndContinue(Level.INFO));
```
LeonardZ0731, line 46: If there is an error with the syncCache, what will the error handler do?

poncem91: This essentially just logs the error and emulates a cache-miss behavior (so it would just return null when an error arises when using get() )

LeonardZ0731: So when the cache doesn't work, we fail to load the user's nickname? Should we have a backup plan for that? I don't think you need to really implement that, but maybe you would like to mention it in the design doc.

poncem91: That's why if the cache value returns null, the nickname is retrieved from datastore instead and pushed into the cache for future use :)

LeonardZ0731: Ah you change it back to retrieving from datastore. I thought you would keep the hashmap with the cache and when the cache doesn't work we would use the hashmap. And then I found you remove the hashmap and I thought you didn't have a backup plan for that. My fault. Then it looks pretty good to me.

